PR #13, opened 2019-06-16
Display lat/long state to user in map window

### LINKED ISSUE #6: Display latitude and longitude on map
Add display of lat/long on map window to show current selection above submit button in bottom right corner of map and allow user to see final lat/long before submitting.

mddilley:
Closes #6


mddilley:
New form displays lat/long state updates from pin drop and also allows lat/long keyboard entry in form
Screen Shot 2019-06-16 at 12 41 10 PM

PR #25, opened 2019-06-20
Satellite tiles

### LINKED ISSUE #9: Change the tiles of the map to satellite tiles
In order to display a more familiar view to user, change the tiles of the map from street view to satellite view.

mddilley:
Closes #9

### DIFF
--- a/src/Components/LayerButtons.js
+++ b/src/Components/LayerButtons.js
@@ -1,0 +1,45 @@
+import React from "react";
+
+const LayerButtons = props => {
+  return (
+    <div className="input-group mb-3 layer-buttons">
+      <div className="input-group-prepend">
+        <div className="ibtn btn-secondary">
+          <label htmlFor="streets-v9" className="input-group-text">
+            Streets
+          </label>
+          <input
+            name="style"
+            id="streets-v9"
+            type="radio"
+            aria-label="Checkbox for following text input"
+            onClick={props.toggleStyle}
+            defaultChecked
+          />
+          <label htmlFor="satellite" className="input-group-text">
+            Satellite
+          </label>
+          <input
+            name="style"
+            id="satellite-v9"
+            type="radio"
+            aria-label="Checkbox for following text input"
+            onClick={props.toggleStyle}
+          />
+          <label htmlFor="satellite-streets-v9" className="input-group-text">
+            Hybrid
+          </label>
+          <input
+            name="style"
+            id="satellite-streets-v9"
+            type="radio"
+            aria-label="Checkbox for following text input"
+            onClick={props.toggleStyle}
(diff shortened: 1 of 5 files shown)


src/index.css
```diff
@@ -14,6 +14,13 @@ code {
 
 /* buttons */
 
+.layer-buttons {
```
mateoclarke: In general, I'd prefer if we had CSS rules imported only into the components that use them instead of at the app level. I think one approach to this is called "CSS modules" and may require some extra webpack config. 
https://flaviocopes.com/react-css/

I'm also open to using CSS-in-JS libraries like [Emotion](https://emotion.sh/docs/introduction) or [Styled Components](https://www.styled-components.com/) or whatever the new cool thing is. I just don't haven't played around with any of these yet. This seems like a nice small project to experiment with.

I say this, but looking back at the mobile signals work orders app I'm seeing my pretty weak CSS patterns in place. So open to any thoughts or ideas here. I've used Sass preprocessor with [BEM](http://getbem.com/) naming conventions a bunch in the past and have really enjoyed that, but I'm not sure how relevant that is these days...

articles: 
- https://blog.bitsrc.io/5-ways-to-style-react-components-in-2019-30f1ccc2b5b
- https://survivejs.com/react/advanced-techniques/styling-react/

src/SelectLocation.js
```diff
@@ -66,10 +67,21 @@ export default class SelectLocation extends Component {
       geocodeAddressString: location.address,
       formValue: props.value || "",
       lat: "Loading...",
-      lng: "Loading..."
+      lng: "Loading...",
+      style: "streets-v9"
```
mateoclarke: I guess this is actually the right place to set the default

src/Components/LayerButtons.js
```diff
@@ -0,0 +1,45 @@
...
+          <label htmlFor="streets-v9" className="input-group-text">
+            Streets
+          </label>
+          <input
+            name="style"
+            id="streets-v9"
+            type="radio"
+            aria-label="Checkbox for following text input"
+            onClick={props.toggleStyle}
+            defaultChecked
```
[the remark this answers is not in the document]
mddilley: 🗺 Done and done. Hybrid set as default and satellite removed.

src/index.css
mddilley: I really like the idea of CSS-in-JS so I went with Styled Components since I haven't had a chance to explore it. I changed the LayerButtons component styles and plan to use this library to clean up the CSS in this project.

src/SelectLocation.js
mddilley: ✅ Map now loads satellite layer on render.

PR #28, opened 2019-06-20
Import Marker component and add test marker to Map

### LINKED ISSUE #18: Display All Submitted Locations on Map
The user will create multiple locations for each work order. All created locations need to display on the map so that the user can verify that they have created all locations correctly. Ideally the map would be zoomed/panned to fit all locations at the largest scale possible.

We'll need to discover how to get lat/lon values for each of the locations either using the existing data that knack requests or by making a new API call.

We also need to consider how the Map centered by default when there are existing locations to add to the map.

We may want to give each map pin a label, like an idea number to differentiate them.

- [x] get knack lat/lon values
- [x] map zoom/scale
- [x] recenter map
- [x] create map pins
- [x] pin labels

mddilley:
Closes #18


src/index.css
```diff
@@ -21,6 +21,11 @@ code {
   margin: 0px 10px 0px 0px;
 }
 
+.popup {
+  font-size: 1rem;
+  font-weight: bold;
+}
+
```
mateoclarke: move this into styled components?

PR #30, opened 2019-06-24
Autozoom to current location

mateoclarke:
_Closes #20_


knack/auto-zoom.js
```diff
@@ -0,0 +1,49 @@
...
+  // Import jQuery into this file from CDN
+  // https://stackoverflow.com/questions/34338411/how-to-import-jquery-using-es6-syntax
+  var script = document.createElement("SCRIPT");
+  script.src =
+    "https://ajax.googleapis.com/ajax/libs/jquery/3.4.1/jquery.min.js";
+  script.type = "text/javascript";
+  document.getElementsByTagName("head")[0].appendChild(script);
+
+  // Create polling function for made sure jQuery is loaded and ready...
+  var checkReady = function(callback) {
```
mddilley: I'd like to talk through this function because I'm not familiar with polling, and I think it might be strategy that could help with a bug in the React side.

package.json
```diff
@@ -11,10 +11,12 @@
     "bootstrap": "^4.3.1",
     "dotenv": "^8.0.0",
     "mapbox-gl": "^0.52.0-beta.2",
+    "mapbox-gl-arcgis-tiled-map-service": "^0.3.0",
```
mddilley: This dependency can be removed from master at some point since it is no longer be used for the aerial layers.

package.json
```diff
@@ -11,10 +11,12 @@
     "bootstrap": "^4.3.1",
     "dotenv": "^8.0.0",
```
mddilley: Note to ask about the appropriate handling of the Mapbox token.

knack/auto-zoom.js
mateoclarke: yeah sounds good!

package.json
mateoclarke: good catch

package.json
mateoclarke: I think we might want to add URL restrictions: https://docs.mapbox.com/help/how-mapbox-works/access-tokens/

mateoclarke: opened an issue for this... #35

PR #36, opened 2019-06-26
Auto update knack fields, remove Knack lat/lon button, and remove lat/lon disp. from React map

### LINKED ISSUE #33: Make lat/lon updates in iframe auto-update the knack location fields
- [x] on lat/lon state change (from search, from dragging, other events?), send a postMessage from react to update the knack location fields
- [x] remove lat/lon fields from react map
- [x] remove get lat/lon button

mddilley:
Closes #33


knack/index.js
```diff
@@ -384,43 +384,29 @@ $(document).on("knack-view-render.view_2587", function(event, scene) {
...
-    iframe.postMessage(message, "*"); //send the message and target URI
-  });
-
   // listen for response
   window.addEventListener("message", function(event) {
     console.log("message received:  " + event.data, event);
-    var latLonResponse = event.data.split(", ");
-    $view_2587.find("form").show();
-    var $latLonFields = $("#kn-input-field_3194");
+    const data = event.data;
```
mateoclarke: I'm on the fence if we should use `const` in knack.js files. It would break the app in IE10<
https://caniuse.com/#feat=const

knack/index.js
```diff
@@ -384,43 +384,29 @@ $(document).on("knack-view-render.view_2587", function(event, scene) {
...
+    if (data.message === "LAT_LON_FIELDS") {
+      var $latLonFields = $("#kn-input-field_3194");
```
mateoclarke: 👏 for prefixing the var with `$`

knack/index.js
```diff
@@ -384,43 +384,29 @@ $(document).on("knack-view-render.view_2587", function(event, scene) {
...
 
-    $latLonFields.find("#latitude").val(latLonResponse[0]);
-    $latLonFields.find("[name='longitude']").val(latLonResponse[1]);
+      $latLonFields.find("#latitude").val(data.lat);
+      $latLonFields.find("[name='longitude']").val(data.lng);
```
mateoclarke: oof, I hate how these fields have weird selectors, wondering if there is anything we could do in the Knack Builder side to configure field IDs. Probably not but I'll check it out

knack/index.js
mddilley: Good to know. I'll use var in the custom Knack JS so that we are using consistent keywords.

knack/index.js
mddilley: 👍

PR #37, opened 2019-06-26
Reorganize map layout in Knack and adjust map style selectors

### LINKED ISSUE #34: Reorganize layout of the map in relation to knack location fields
- [ ] Once we've removed the lat/lon fields from displaying inside the React iframe map, we need to make the default Knack lat/lon fields work better with the layout of the map.

- [x] We also need to bump down tile selector a hair not to overlap with the search field.

mddilley:
Closes #34


mddilley:
@mateoclarke The jQuery to overlay the lat/lon fields and button in the Knack app is added to branch 32_migrate since the Knack custom JS is referencing that branch in the builder. Master is pulled into this branch too.

PR #52, opened 2019-07-01
Update map on delete of location

### LINKED ISSUE #40: Update map when location is deleted
- might need to trigger a page refresh, or re-render of iframe, or send a new message to remove the deleted location ID from the map.

mddilley:
Closes #40


mddilley:
Map reloads and updates markers on deletion and addition of locations in the Knack app. Format of the lat/lon sent from the React app to Knack to populate the Knack form fields was changed to 7 decimal places.

PR #61, opened 2019-07-02
Show React app map in Location Details modal

### LINKED ISSUE #41: Add map to Location Details modal
- [x] remove Here map (we could do this w/ jQuery or in the Knack builder
- [x] replace with Mapbox
- [x] plot current location in one color, with other signs also on the map in another color

mddilley:
Closes #41


mddilley:
@mateoclarke This branch is working, but it is blocked by PR #57 since it requires the app to handle one location. New custom JS for Knack lives in addMapToLocationDetails.js in this branch (saved in case either of us needs to replace the Knack JS for a separate issue).

mddilley:
@mateoclarke Just ran into another blocker. Since the Location Details view is a modal, the messages being sent to and from the React app are generating changes in the first iFrame rendered (beneath the modal) and not the second iFrame map (within the modal).

PR #71, opened 2019-07-05
Change map marker shadow color to red

### LINKED ISSUE #59: Increase Map Marker Contrast
It can be difficult to see the gray map marker when placing a location on the map, especially with the satellite basemap. Can this marker be changed to a style that will contrast more with the basemap?

mddilley:
Closes #59 
Screen Shot 2019-07-05 at 9 49 09 AM

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,8 @@
     "start": "react-scripts start",
     "build": "react-scripts build",
     "test": "react-scripts test",
-    "eject": "react-scripts eject"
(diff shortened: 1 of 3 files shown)


johnclary:
@mddilley this looks good, but we may need to change the color of the marker itself; i find it hard to see sometimes. but we can hold off for more feedback from the users. this is fine w/ me to merge

PR #76, opened 2019-07-06
Improve layout of lat/lon fields and "Add location" button

### LINKED ISSUE #51: Make the default Knack lat/lon fields work better with the layout of the map
Currently, the lat/lon fields and "Add Location" buttons sit below the iFrame React map in the Knack app. These fields and buttons should be styled to overlay the map.

The jQuery below can be used, but these style changes occur each time a location is added or deleted (leading them to slowly move away from the map). Need to address keeping them static (overlaying the map in all cases).

$("#view_2607 button").css({marginTop: '-=125px'});
$("#view_2607").css({marginTop: '-=130px'});
$("#view_2607 > div.view-header > p").hide();
$("#view_2607 button").css({marginLeft: '+=5px'});
$("#view_2607").css("color", 'white');
$("#view_2607").css("textShadow", '1px 1px 1px rgba(0, 0, 0, 1)');

mddilley:
Closes #51

### DIFF
--- a/knack/index.css
+++ b/knack/index.css
@@ -1,0 +1,43 @@
+/* hide attachment type column (field contents are extracted via JS */
+div.view_2107 .field_2403 {
+  display: none;
+}
+div.view_2108 .field_2403 {
+  display: none;
+}
+
+/* hide attachment type columns (in Traffic Counts), field contents extracted via JS */
+div.view_2357 .field_3174 {
+  display: none;
+}
+div.view_2486 .field_3174 {
+  display: none;
+}
+
+/* hide attachment type columns (in Manage Requests under Traffic Counts), field contents extracted via JS */
+div.view_2491 .field_3174 {
+  display: none;
+}
+
+/* hide attachment type columns (in Request Status under Traffic Counts), field contents extracted via JS */
+div.view_2465 .field_3174 {
+  display: none;
+}
+
+/* move and style lat/lon fields and button to overlay React app map */
+#lat-lon-form {
+  color: white;
+  text-shadow: 1px 1px 1px rgba(0, 0, 0, 1);
+  margin-top: -130px;
+}
+
+#lat-lon-form button {
+  margin-top: -125px;
+  margin-left: +5px;
+}
+
+#lat-lon-form label.help {
+  font-weight: bold;
+  font-size: 13px;
+  margin-top: 1px;
+}
--- a/knack/index.css
+++ b/knack/index.css
@@ -24,20 +24,34 @@ div.view_2465 .field_3174 {
   display: none;
 }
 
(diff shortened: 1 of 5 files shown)


mddilley:
@mateoclarke That issue with the lat/lon fields and button shifting after submitting a location was due to a header div inserting into the DOM after submission. After removing that div with jQuery upon render of that form's view, the fields/button remain on the map when adding or deleting locations. 
Screen Shot 2019-07-06 at 4 58 11 PM

knack/index.js
```diff
@@ -531,28 +531,22 @@ $(document).on("knack-view-render.view_2609", function(event, scene) {
...
 
-// IN PROGRESS: Mike's code for 51_lat_lon_layout
-$(document).on("knack-view-render.view_2607", function(event, scene) {
-  $header = $("#view_2607 > div.view-header");
-  $(document).on("knack-view-render.view_2586", function(event, scene) {
-    $("#view_2586").append($header);
-  });
-  $form = $("#view_2607");
+// Overlay Latitude/Longitude fields and button on map
+$(document).on("knack-scene-render.scene_1028", function(event, scene) {
```
mateoclarke: Looks like we're using the `view-render` event in addition to the `scene-render`. Does the code for each of these things need to be in separate places?

knack/index.css
```diff
@@ -0,0 +1,43 @@
...
+div.view_2491 .field_3174 {
+  display: none;
+}
+
+/* hide attachment type columns (in Request Status under Traffic Counts), field contents extracted via JS */
+div.view_2465 .field_3174 {
+  display: none;
+}
+
+/* move and style lat/lon fields and button to overlay React app map */
```
mateoclarke: I wonder if it would be nice, for the sake of documentation, to add more about where these classes are used.

mateoclarke:
@mddilley I didn't open this up to preview in Knack, but just from your screenshot, it would nice if the lat & lon fields had the same width.

knack/index.js
mddilley: The jQuery in the view-render does need to be separate from the scene-render since that view renders after the submission of the lat/lon fields without a full page reload (the jQuery inside the scene-render handles all of the changes for full page reload). There may be a way to combine these, but it gets tricky with the form submission.

knack/index.css
mddilley: Agreed - it would be very hard to find these targets after moving on from this issue. I added the name of the page and scene id to the comment. Let me know if any other info would help, and I'll add.

mddilley:
@mateoclarke  I adjusted the width of the lat field and move the lon field over to the left. I also moved the labels above the fields and the button below the field.
Screen Shot 2019-07-08 at 1 49 28 PM

PR #90, opened 2019-07-09
Remove Unused Code from Data Tracker Migration

johnclary:
We copied this app from the Data Tracker, so there is a bunch of code specific to that app that we do not need here.


mateoclarke:
@johnclary is this ok to merge? I need to make a change to the knack/index.js file and I was working from master.

mateoclarke:
@johnclary we know have a conflict to resolve since @dianamartin committed direct to `master`. Can you reconcile and update the Knack.js file? Also note that I also have an open PR whose code had been updated in Knack, not sure if Diana's changes overwrote those or note.

mateoclarke:
merged my code into here with the merge conflict resolution for master. going to merge now so this doesn't get stale.

PR #97, opened 2019-07-10
Inventory proper map rendering by role

mateoclarke:
_Closes #88_

This PR is attempting to make sure we have properly inventoried all view IDs were maps belong with considerations for both the Technician/Editing role (Test Technician) & the Viewer role (Test Viewer)

## Work Order Details Page - Editable - view_2573
Screen Shot 2019-07-10 at 6 52 09 PM

## Work Order Details Page - Viewable - view_2619
_NOTE: on this one, we shouldn't be displaying the Red select location pin, as we currently are. We should only see the fixed pins. Need to figure out a work around here that doesn't break the editor view_
Screen Shot 2019-07-10 at 6 46 20 PM

## Location Details - Editable - ~view_2609~ view_2733  
View 2609 is the Edit button. That only shows in the Editable mode. Both the edit and view Locations Details pages have view_2733. So we can use the same viewId for both technician and viewer roles in this case.

Screen Shot 2019-07-10 at 6 56 15 PM
 
## Location Details - Viewable - view_2733  

Screen Shot 2019-07-10 at 6 33 21 PM

## Edit Location - view_2682

Screen Shot 2019-07-10 at 6 56 45 PM

### DIFF
--- a/knack/iframeMapMessenger.js
+++ b/knack/iframeMapMessenger.js
@@ -34,7 +34,7 @@
     // Add React app as iframe if iframe doesn't already exist
     if ($(myView + " #mapIFrame").length === 0) {
       $(
-        '<iframe src="https://5d2115b19ceaed0007c6ae7d--atd-geo-knack-ui.netlify.com/" frameborder="0" scrolling="yes" id="mapIFrame" \
+        '<iframe src="https://atd-geo-knack-ui.netlify.com/" frameborder="0" scrolling="yes" id="mapIFrame" \
     style="width: 100%;height: 523px;"></iframe>'
       ).appendTo($viewSelector);
     }
--- a/knack/iframeMapMessenger.js
+++ b/knack/iframeMapMessenger.js
@@ -58,7 +58,6 @@
     });
 
     // Location Details Page Maps
-
     function locationDetailsMapMessage(viewId) {
       var locationViewIFrame = $("#" + viewId + " #mapIFrame")[0].contentWindow;
       var urlArray = window.location.href.split("/");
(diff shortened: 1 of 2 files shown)


mateoclarke:
@mddilley, need your help thinking through a solution for the select location pin rendering on the Work Order Details Page (Test Viewer/view_2619). Instead, we'd want that map to only show the fixed pins. We can save that for another PR.

knack/iframeMapMessenger.js
```diff
@@ -76,12 +77,23 @@
       };
 
       sendMessageToApp(markerMessage, locationViewIFrame);
+    }
+    // Location Details Page - Editable
+    $("#view_2609 #mapIFrame").on("load", function() {
```
mddilley: Does this need to happen after the change to rendering the map in view_2733 in both Editable and Viewable Location Details views? I'm glad you made that switch btw since it makes the view consistent across roles. 👍

knack/iframeMapMessenger.js
```diff
@@ -76,12 +77,23 @@
...
     });
 
-    $("#mapIFrame").on("load", function() {
+    // Location Details Page - Viewer
+    $("#view_2733 #mapIFrame").on("load", function() {
+      locationDetailsMapMessage("view_2733");
+    });
+
+    // Work Orders Details Page Maps
+    function workOrdersDetialsMapMessage(viewId) {
```
mddilley: Small typo here in the function name (detials instead of details) here and in the calls on line 111 and 115. Doesn't affect the function.

mateoclarke:
@mddilley @johnclary @dianamartin Because of the active changes happening in the master branch, I'm going to go ahead and merge this branch into John's so we can get back in sync.

PR #102, opened 2019-07-11
Fix map markers not showing in initial render

### LINKED ISSUE #98: Map markers don't show on initial map render
In some cases I've noticed the map markers don't show until you change the zoom or move the map.

At first:
Screen Shot 2019-07-10 at 6 33 31 PM

After moving/zooming:
Screen Shot 2019-07-10 at 6 33 21 PM

mddilley:
Closes #98

### DIFF
--- a/knack/index.js
+++ b/knack/index.js
@@ -492,7 +492,7 @@ $(document).on("knack-view-render.view_2893", function(event, page) {
   bigButton(
     "markings-gis-qa",
     "view_2893",
-    "https://atd.knack.com/signs-markings#marking-gis-qa/",
+    "https://atd.knack.com/signs-markings#markings-gis-qa/",
     "road",
     "GIS QA | Markings"
   );
--- a/knack/index.js
+++ b/knack/index.js
@@ -394,9 +394,9 @@ $(document).on("knack-view-render.view_2733", function(event, scene) {
 
 // Overlay Latitude/Longitude fields and button on map
 $(document).on("knack-scene-render.scene_1028", function(event, scene) {
-  // Separate header from form and move form below map
+  // Remove header from form to prevent lat/lon fields from shifting and move form below map
   var $header = $("#view_2607 > div.view-header");
-  $("#view_2586").append($header);
+  $header.remove();
   var $form = $("#view_2607");
   $form.attr("id", "lat-lon-form");
   $form.detach();
--- a/src/SelectLocation.js
+++ b/src/SelectLocation.js
@@ -553,12 +553,7 @@ export default class SelectLocation extends Component {
                   </Marker>
                 ))}
               {viewLocation.length !== 0 && (
-                <Marker
-                  key={sign.id}
-                  anchor="bottom"
-                  coordinates={[sign.lng, sign.lat]}
-                  onClick={() => this.signClick(sign.id)}
-                >
+                <Marker anchor="bottom" coordinates={viewLocation}>
                   <img src="/red-icons8-marker-40.png" />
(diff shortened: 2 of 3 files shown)


mddilley:
@mateoclarke The markers now load as soon as the map renders. I did this by changing the components from Layers w/ Features to Markers only. Not as performant but documentation says it only leads to slowdown when there are a large # of markers.

This branch has a few changes unrelated to #98 
- Made url correction for @dianamartin in index.js
- Removed header div that was shifting the lat/lon fields and button off the map in Knack app
- Disabled re-center map on Marker click (led to overlap with the main red pin and pulse shadow)

knack/index.js
```diff
@@ -394,9 +394,9 @@ $(document).on("knack-view-render.view_2733", function(event, scene) {
 
 // Overlay Latitude/Longitude fields and button on map
 $(document).on("knack-scene-render.scene_1028", function(event, scene) {
-  // Separate header from form and move form below map
+  // Remove header from form to prevent lat/lon fields from shifting and move form below map
   var $header = $("#view_2607 > div.view-header");
-  $("#view_2586").append($header);
+  $header.remove();
```
mateoclarke: 👽 🔫

src/SelectLocation.js
```diff
@@ -112,10 +118,10 @@ export default class SelectLocation extends Component {
   signClick = id => {
     // set state.sign for Popup parameters in render, center map to clicked sign
     const clickedSign = this.state.signs.find(sign => sign.id === id);
-    const newCenter = [clickedSign.lng, clickedSign.lat];
+    // const newCenter = [clickedSign.lng, clickedSign.lat];
     this.setState({
-      activeSign: clickedSign,
-      center: newCenter
+      activeSign: clickedSign
+      // center: newCenter
```
mateoclarke: I didn't understand what this changed from your comment but I trust your judgement if you think its an improvement. Send a screenshot of what changed if you want more feedback on it.

Let's go ahead and delete the commented out lines once you feel good about it.

src/SelectLocation.js
```diff
@@ -546,30 +541,28 @@ export default class SelectLocation extends Component {
                   <div className="pulse" />
                 </>
               )}
-
-              <Layer type="symbol" id="signs" layout={layoutLayer}>
-                {signs.map(sign => (
-                  <Feature
+              {signs &&
+                signs.map(sign => (
+                  <Marker
```
mateoclarke: Remind me why Marker component work better than Features in a Layer? You explained how it isn't as performant but I guess I still don't understand how did it fix the problem? Stack Overflow said so is an ok answer.

src/SelectLocation.js
mddilley: The old behavior was that the map would re-center on a Marker when it was clicked to see the details pop-up. It felt jarring after using it for a while, and it could get annoying on mobile since the map window can feel smaller (easier to lose your place). The new behavior is the map does not re-center when a Marker is clicked.

src/SelectLocation.js
mddilley: Documentation says that Features in a Layer should work just fine so I tried refactoring the logic for rendering Layers. I was never able to have the Features load immediately so I think there may be something more complex going on. I decided to try Markers since it is a simpler implementation, and it seems more appropriate for the number of Markers this app will drop. Luckily, that fixed the rendering issue.

I'd like to get to the bottom of it since there are some really cool demos here with these types of features working great. https://alex3165.github.io/react-mapbox-gl/demos

PR #104, opened 2019-07-12
Disable red pin for Viewer Work Order Details page

### LINKED ISSUE #103: Remove red select location pin on Work Order Details Page - Viewable - view_2619
On this one, we shouldn't be displaying the Red select location pin, as we currently are. We should only see the fixed pins. Need to figure out a work around here that doesn't break the editor view

Screen Shot 2019-07-10 at 6 46 20 PM

mddilley:
Closes #103

### DIFF
--- a/knack/iframeMapMessenger.js
+++ b/knack/iframeMapMessenger.js
@@ -34,7 +34,7 @@
     // Add React app as iframe if iframe doesn't already exist
     if ($(myView + " #mapIFrame").length === 0) {
       $(
-        '<iframe src="https://localhost:9001/" frameborder="0" scrolling="yes" id="mapIFrame" \
+        '<iframe src="https://atd-geo-knack-ui.netlify.com/" frameborder="0" scrolling="yes" id="mapIFrame" \
     style="width: 100%;height: 523px;"></iframe>'
       ).appendTo($viewSelector);
     }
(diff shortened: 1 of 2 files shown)


knack/iframeMapMessenger.js
```diff
@@ -34,7 +34,7 @@
     // Add React app as iframe if iframe doesn't already exist
     if ($(myView + " #mapIFrame").length === 0) {
       $(
-        '<iframe src="https://atd-geo-knack-ui.netlify.com/" frameborder="0" scrolling="yes" id="mapIFrame" \
+        '<iframe src="https://localhost:9001/" frameborder="0" scrolling="yes" id="mapIFrame" \
```
mateoclarke: I don't think we want this referencing localhost in prod right @mddilley?

mateoclarke: Right now in the knack js code (index.js), it references the "https://dnb4pix4gcpf6.cloudfront.net/atd-knack-signs-markings/98_marker_render/iframeMapMessenger.js". Do we need to update that?

src/SelectLocation.js
```diff
@@ -368,6 +369,10 @@ export default class SelectLocation extends Component {
 
       switch (data.message) {
         case "SIGNS_API_REQUEST":
+          // Use workOrderDetailsViewable state to disable red pin and pulse if user is Viewer role
+          if (data.view === "view_2619") {
```
mateoclarke: Nice, this is pretty clever. Kinda of a hard-coded hack but let's count our wins and never talk about this again.

knack/iframeMapMessenger.js
mddilley: Good catch! Just changed and pushed. I'll update the Knack JS in builder with the latest from master after merging this branch.

PR #139, opened 2019-07-23
114 Attachments

### LINKED ISSUE #114: Attachments JS and CSS
The attachments JS and CSS need to be added to the application. 

Right now you see the whole file name rather than the attachment type
image

Should look like this:
image

@johnclary Would you be able to do this?

mateoclarke:
Closes #114 

Adds attachments to this one place updating the CSS & JS

Screen Shot 2019-07-22 at 10 37 47 PM


dianamartin:
@mateoclarke The attachments look good :+1:

PR #167, opened 2019-08-01
Use Here geocoder results for intersections

### LINKED ISSUE #162: Implement using @ in intersection geocoding in address bar in iFrame map
- [x] Check if geocoding is currently using this option
- [ ] Implement to improve user search for street intersections

Link from Sergio https://developer.here.com/api-explorer/rest/geocoder/latitude-longitude-intersection

mddilley:
Closes #162

### DIFF
--- a/src/SelectLocation.js
+++ b/src/SelectLocation.js
@@ -139,7 +139,8 @@ export default class SelectLocation extends Component {
       .get(
         `https://geocoder.api.here.com/6.2/geocode.json?bbox=30.627918,-98.173053;30.02329,-97.369564&city=Austin&street=${firstStreet +
           " @ " +
-          secondStreet}&`
+          secondStreet}&app_id=J6lGCnJcHK0WL5HbHQMw&app_code=vUvXCuVInzjjThg2H6UiQg
+          &gen=9`
       )
       .then(res => {
         const location = res.data.Response.View[0].Result[0].Location;
--- a/src/SelectLocation.js
+++ b/src/SelectLocation.js
@@ -27,7 +27,6 @@ const geocoderControl = new MapboxGeocoder({
   // countries: 'us',
   trackProximity: true,
   limit: 5
-  // localGeocoder: forwardGeocoder
 });
 
 const geolocateControl = new GeolocateControl({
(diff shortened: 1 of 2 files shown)


mddilley:
This PR adds intersection results from the Here Geocoder API to the Mapbox API results that populate below the address field. The search requires that the user enters "first street and second street" (ex. "45th and lamar"). The result from Here populates with a pin emoji to the left to indicate an intersection (this can be removed or updated - I used it to differentiate results while developing).

@mateoclarke The only question to close this out is how to handle the Here app id and app code needed in the url. @sergio how have you been handling these? I've left them out of my commits on this branch to keep them secret.

**Search/autopopulate**
Screen Shot 2019-08-01 at 1 06 09 PM

**Location selected**
Screen Shot 2019-08-01 at 12 52 46 PM

sergiogcx:
Thanks for keeping the app_id and and app code secret, since my ETL process runs on a server, I secure them via environment variables. I know ODD uses them out in the open: https://github.com/cityofaustin/us-forms-system/blob/2ca446d2367db68644f1994026367e4bfc57a2ea/src/js/widgets/SelectLocation/SelectLocationWidget.jsx#L7

It is possible to protect the keys behind server-side rendered pages, but then we would have to talk about how we protect that endpoint too.

Let's have a conversation.

mddilley:
I added handling of "&" and "@" delimiters along with "and" as we discussed with @dianamartin . The new Here credentials are added too so this is good to go.

dianamartin:
@mddilley We applied these changes to Signs and Markings PROD, correct?

mddilley:
@dianamartin This code will go live when this PR is reviewed and then merged so it isn't live in PROD yet.

dianamartin:
Thanks for the deets!

src/SelectLocation.js
```diff
@@ -87,7 +77,8 @@ export default class SelectLocation extends Component {
       initialLoad: false,
       zoom: "",
       viewLocation: [],
-      workOrderDetailsViewer: false
+      workOrderDetailsViewer: false,
```
mateoclarke: Not a big deal because its not related to this PR that much but the naming convention in JS for boolean typed variables is to start with `is` or `has`.  Here seems like `isWorkOrderDetailsViewer` would make the most sense.

https://www.samanthaming.com/tidbits/34-better-boolean-variable-names

mddilley: Good to know - I like learning about self-documenting code and this isn't a convention that I was familiar with. Thanks!

PR #205, opened 2019-08-28
Remove Data Tracker-specific JS

johnclary:
Removing legacy JS from the Data Tracker which is not relevant to the signs + markings app.


mateoclarke (approved):
👍

Issue #209, opened 2019-08-29
Signs Work Orders: Hide Lat/Lon Fields


johnclary:
There's not much screen real estate when viewing signs work orders on a tablet. Hide the lat/lon fields from the embedded map on the Sign Work Order Details. ![Screen Shot 2019-08-28 at 10.29.05 PM.png](https://images.zenhubusercontent.com/5caf676becad11531cc417cb/ecad3d1c-6277-4d4a-84f8-866e6cfc5b73)

- [x] Hide the `Latitude` and `Longitude` fields on the map
- [x] Make the `Add Location` button larger` #208

johnclary:
@mateoclarke @mddilley can we get this done during the current sprint?

mateoclarke:
Yup, we'll knock it out

mddilley:
Making map submit button larger addressed in PR #215

PR #215, opened 2019-08-29
Increase Form Submit Button sizes

### LINKED ISSUE #208: Increase Form Submit Button Sizes for Mobile Devices
After testing on the 8" tablets, we need to increase the form submit button sizing for mobile users.

Here's the `button` element of concern. We will need to isolate this change to `type="submit"`:
```
<button class="kn-button is-primary" type="submit">
      Submit
    </button>
```

Ideally, the button would be at least twice the current height, with text size increased as well, and full-width.

mddilley:
Closes #208

### DIFF
--- a/knack/index.css
+++ b/knack/index.css
@@ -116,15 +116,17 @@ a.small-button {
   font-weight: 900;
 }
 
-/* Make buttons with type="submit" bigger */
-button[type="submit"] {
-  height: 64px !important;
-  font-size: 32px !important;
-  min-width: fit-content;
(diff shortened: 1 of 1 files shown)


mddilley:
This PR doubles size and font size and widens buttons with `type="submit"` in the app. This PR also addresses the map submit button part of #209.

## Bigger buttons
Screen Shot 2019-08-29 at 11 30 28 AM

## Map overlay button
Screen Shot 2019-08-29 at 11 45 23 AM

PR #216, opened 2019-08-29
Hide latitude/longitude fields

### LINKED ISSUE #209: Signs Work Orders: Hide Lat/Lon Fields

mddilley:
This PR closes #209. The lat/lon fields were hidden and the styles for the hidden elements were removed. The submit button shifted slightly after hiding the other form elements so the margin-top was adjusted to keep it in place.

### Map without lat/lon fields
Screen Shot 2019-08-29 at 12 20 29 PM

### DIFF
--- a/knack/index.css
+++ b/knack/index.css
@@ -116,15 +116,17 @@ a.small-button {
   font-weight: 900;
 }
 
-/* Make buttons with type="submit" bigger */
-button[type="submit"] {
-  height: 64px !important;
-  font-size: 32px !important;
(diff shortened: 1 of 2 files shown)


mateoclarke (approved):
👍

PR #226, opened 2019-09-06
Update styling on Add Location button

### LINKED ISSUE #218: Signs Work Orders: Increase "Add Location" Button Contrast
@johnclary commented on [Wed Aug 28 2019](https://github.com/cityofaustin/atd-data-tech/issues/534)

It's hard to see the "Add Location" button from the map on the Signs Work Order details page. Let's increase the contrast to optimize for use outdoors. E.g., White button with black text.

mateoclarke:
Closes #218 

Screen Shot 2019-09-06 at 4 29 35 PM

### DIFF
--- a/knack/index.css
+++ b/knack/index.css
@@ -37,9 +37,9 @@ div.view_2465 .field_3174 {
 
 #lat-lon-form button {
   margin: -10px 0 0 10px;
-  color: black !important;
+  color: #07467c !important;
   background: white !important;
-  border: 3px solid black;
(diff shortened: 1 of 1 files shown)


dianamartin:
The button definitely is hard to miss now. I would say if we could just change the black outline and font to the navy blue we had then I think we're done here.

mateoclarke:
@dianamartin updated the color, going to merge...

Screen Shot 2019-09-06 at 4 45 35 PM

Issue #214, opened 2019-08-29
Increase default menu button size


dianamartin:
## Request to increase the height on the default menu buttons.
Signs and Markings technicians would regularly use these buttons and on a tablet, these would be great if they were larger buttons to click on.

### [Signs - My Work Orders](https://atd.knack.com/signs-markings#my-work-orders/)
`scene: 1085`

![image](https://user-images.githubusercontent.com/21112975/63952408-2a22f900-ca45-11e9-8c41-2aabdeda26e1.png)

### [Signs - Work Order Details](https://atd.knack.com/signs-markings#my-work-orders/view-work-orders-details-sign/5d39f42940c00f24eaf9e05a/)
`scene: https://builder.knack.com/atd/signs-markings#pages/scene_1028/views/view_2684`

![image](https://user-images.githubusercontent.com/21112975/63952998-1b891180-ca46-11e9-8c02-cf893fa3b44c.png)

### [Markings - Jobs](https://atd.knack.com/signs-markings#jobs/)
`scene: 717`

![image](https://user-images.githubusercontent.com/21112975/63952563-6f472b00-ca45-11e9-9298-e2d02b0f9306.png)

### [Markings - Job Details](https://atd.knack.com/signs-markings#jobs/view-work-orders-markings-job-details/5d3a285cafee63251a093a1d/)
`scene: 716`

![image](https://user-images.githubusercontent.com/21112975/63952764-bfbe8880-ca45-11e9-9c03-e5482d059754.png)

`scene: 894`
![image](https://user-images.githubusercontent.com/21112975/63952868-e977af80-ca45-11e9-9c01-8829c2ca9648.png)

mateoclarke:
@dianamartin ~I don't see the buttons for "Signs - Work Order Details". Do you have to be logged in with certain user roles to see buttons? If so, what users and can you DM me this details.~ nvm, I updated my user's roles and I see them now.

dianamartin:
### Modifications in TEST to fit tablet size view
**Signs Work Order Details** page
`Technician menu` buttons
- Attachment (CHANGED, removed "Add")
- Comment (CHANGED, removed "Add")
- Edit
- Submit (CHANGED, removed "for Review")

**Jobs** page
`Jobs by Work Group` buttons
- Long line
- Short line
- RPM
- Specialty (CHANGED, removed "Markings")
- [x] Removed export button
- [ ] Remove key word search?

**Job Details** page
`Technician menu` buttons
[...]

Issue #213, opened 2019-08-29
Hide "Back to My Work Orders | Assigned" and Back to top button


dianamartin:
On Work Order Details page, the the very bottom there is a link "Back to My Work Orders | Assigned". Signs and Markings would like to hide link from technicians view because they accidentally click on this link instead of going back to the top of the page to click "Submit for Review".

Additionally, adding a "back to top" button would be very useful as well so that the technician can just click it and it will return the user to the top of the page to see the work order buttons. 

![image](https://user-images.githubusercontent.com/21112975/63952076-820d3000-ca44-11e9-9149-5702ebb24d83.png)

PR #227, opened 2019-09-06
Promote button size, back to top, and add location btn contrast updates to production

mateoclarke:
Includes code from #213 #214 #218


dianamartin:
Slack convo
>I just refreshed, I like it!

PR #293, opened 2019-12-10
Post-Migration Cleanup

johnclary:
Remove leftover Data Tracker JS and CSS which is not needed in this dedicated Signs & Markings app.

### DIFF
--- a/knack/index.css
+++ b/knack/index.css
@@ -9,24 +9,6 @@ div.view_2565 .field_2403 {
   display: none;
 }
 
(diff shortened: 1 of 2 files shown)


johnclary:
FYI @dianamartin and @SurbhiBakshi I tested and deployed these changes in S&M prod. There should be no issues, but if the users report issues tomorrow, we'll want to look back at these changes.

johnclary:
Also @dianamartin @SurbhiBakshi if you have changes to the javascript or css in a signs & markings test environment, we'll need to review those changes before updating prod. We can use github for this. Let me know.

PR #317, opened 2021-04-29
Knack iFrame | Update Netlify url to avoid redirect through http address

mddilley:
This PR updates the source url for the iFrame map that embedded in the Knack Signs and Markings app. After Netlify updated their url format, a redirect started occurring as shown in the screenshot below. Not sure why this didn't surface earlier, but the only difference that I can think of is the migration to the new Knack builder.

The error in the Knack app
Screen Shot 2021-04-29 at 10 02 46 AM

The redirect happening when accessing https://atd-knack-signs-markings.netlify.com/
Screen Shot 2021-04-29 at 10 29 16 AM

### DIFF
--- a/knack/iframeMapMessenger.js
+++ b/knack/iframeMapMessenger.js
@@ -34,7 +34,7 @@
     // Add React app as iframe if iframe doesn't already exist
     if ($(myView + " #mapIFrame").length === 0) {
       https: $(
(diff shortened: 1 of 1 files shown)


sergiogcx (approved):
🚢

mateoclarke (commented):
🚢

PR #324, opened 2025-04-01
Rebuild the signs work order map with mapbox

chiaberry:
Collector branch for https://github.com/cityofaustin/atd-data-tech/issues/21517

I copied most of the map and the geocoder from the vision zero crash map. react-mapbox-gl is at version `^7.1.9` in vision zero, but [the package](https://visgl.github.io/react-map-gl/docs) say its on v8 now, so the [imports are different](https://visgl.github.io/react-map-gl/docs/upgrade-guide) and the types will be different. 

```bash
cd signs-work-order-map
```
then to first spin up, `nvm use` and `npm install` and `cp env_template .env`, fill out the mapbox token. 

```bash
npm run dev
```

You should see a full screen map. Test putting in an address in the upper left input. it should take you to that address
You can drag the map and the marker will recenter and i am console.logging the new lat/lon

For comparison, here is the existing map deployed https://atd-knack-signs-markings.netlify.app/

Does this feel like a good start to a collector branch?


chiaberry:
Something I completely overlooked when reviewing this repo last month was the [calcs file](https://github.com/cityofaustin/atd-knack-signs-markings/blob/staging/src/calcs.js), which renders some calculation widgets https://atd-knack-signs-markings.netlify.app/calcs, embedded in a page in knack. Christina is checking if this is still being used by SMD and if it needs to be included when we redeploy.

signs-work-order-map/.nvmrc
```diff
@@ -0,0 +1 @@
+20
```
johnclary: I _think_ can we go all the way up to `22`... at least `22` worked for my testing

signs-work-order-map/app/globals.css
johnclary, line 1: would be nice to use `scss` from the get-go. i believe you can simply `npm install sass` and change the file extension.

signs-work-order-map/app/page.module.css
johnclary, line 1: is the plan to use bootstrap, eventually?

signs-work-order-map/app/page.tsx
```diff
@@ -0,0 +1,60 @@
+"use client";
```
johnclary, line 1: I'm still fuzzy on when/how to avoid client-only pages, but my vague understanding is that the further you push client-only components down your tree, the more your site is server-side/cacheable/etc. 

we get zero benefit from this in VZE because the entire app uses the client-only Auth0 provider 😞 .

 not saying that's worth it yet (after all, the entire app is a JS webmap), but just want to keep this consideration in mind.

signs-work-order-map/app/page.tsx
```diff
@@ -0,0 +1,60 @@
...
+  DEFAULT_MAP_PAN_ZOOM,
+  MAP_COORDINATE_PRECISION,
+} from "@/config/map";
+
+interface LatLon {
+  latitude: number;
+  longitude: number;
+}
+
+export default function Home() {
```
johnclary, line 60: assuming this page is going to grow, might be good to move the map into it's own component now?

signs-work-order-map/env_template
johnclary, line 1: for consistency with VZE, should we save our vars to `.env.local` instead of `.env`? in earlier versions of nextjs i thought there were some important benefits from using `.env.local` vs `.env` for local dev, but as far as i can tell this is more of a consideration when using Vercel for deployments.

signs-work-order-map/package.json
```diff
@@ -0,0 +1,31 @@
...
+  "dependencies": {
+    "@mapbox/mapbox-gl-geocoder": "^5.0.3",
+    "@types/mapbox__mapbox-gl-geocoder": "^5.0.0",
+    "@types/mapbox-gl": "^3.4.1",
+    "mapbox-gl": "^3.10.0",
+    "next": "15.2.4",
+    "react": "^19.0.0",
+    "react-dom": "^19.0.0",
+    "react-map-gl": "^8.0.1",
+    "react-map-gl-geocoder": "^2.2.0"
```
johnclary: this package's dependencies seems to be responsible for a handful of vulnerability warnings. i see we are instead using `@mapbox/mapbox-gl-geocoder` in the VZE. that's because of the difference in `react-map-gl` versions?

signs-work-order-map/.eslintrc.json
```diff
@@ -0,0 +1,10 @@
+{
+  "extends": ["next/core-web-vitals", "next/typescript", "prettier"],
```
johnclary, line 2: when i run `npm run lint` i get an error about a missing prettier config 💥

signs-work-order-map/app/page.tsx
```diff
@@ -0,0 +1,60 @@
...
+  const onMoveEnd = useCallback((e: ViewStateChangeEvent) => {
```
johnclary: observing that the VZE listens to `onDrag` so that the pin stays centered while the map slips around. it'd be cool to to test out that fancier implementation eventually, although i'm sure it requires more CPU. this has me thinking that'd it be nice to be know the devices that are going to be used for this and perhaps even fire up good ole browser stack for occasional testing 🤔

johnclary (commented):
This looks so good! Thanks @chiaberry for getting us set up with scaffolding to rebuild this app 🚀 🚀 🚀

signs-work-order-map/app/page.tsx
chiaberry: oooh browserstack! yes. and +1 on knowing what devices are used, which tablets etc.

signs-work-order-map/app/page.module.css
chiaberry: I thought we would use bootstrap for sure if we needed to bring the materials calculators into this app, but otherwise I wasn't sure. The existing app uses bootstrap but I think only for that page.

signs-work-order-map/app/page.tsx
chiaberry: I included this because next kept giving me an error saying it was needed, and if you look at an old commit i had the comment "why is next telling me to use this??"

signs-work-order-map/package.json
mddilley: changes they've made over the last few versions make this confusing, but I think that `react-map-gl-geocoder` can be removed. their latest geocoder example doesn't show it https://visgl.github.io/react-map-gl/examples/mapbox/geocoder. also just cool to see that they're on to version 8!

signs-work-order-map/.eslintrc.json
mddilley: i see this too. i think adding the eslint plugin like is in the VZE dev dependencies will get this back in business 😎

signs-work-order-map/app/globals.css
mddilley: that would be a nice way to keep this consistent between apps too. iirc there is some need to style map buttons that this could make easier

signs-work-order-map/app/layout.tsx
```diff
@@ -0,0 +1,20 @@
+import type { Metadata } from "next";
+import "./globals.css";
+
+export const metadata: Metadata = {
+  // todo: what should this say
+  title: "Signs and Markings Operations",
+  description: "City of Austin Transportation Public Works",
```
mddilley: you mean you don't want it to say "React App" like the other one? j/k 😅 This all looks reasonable to me! maybe adding the "and" between Transportation Public Works to be super official.

signs-work-order-map/app/page.module.css
mddilley: cool, that makes sense to me to hold out on Bootstrap to see if the calculator code is needed. Is anything besides lines 1-14 doing anything? can we remove? 

i noticed styles in the media query on line 34 are causing the map to shift down and leave blank space at mobile width. probably won't affect tablets, but I think the main concern is getting this map to cover the entire embedded iframe no matter what.

signs-work-order-map/app/page.tsx
mddilley: yeah, this is tricky. i read a thing about server components once that stuck with me that described server components as the static branches that lead down to the leaves which are client components that need client-side interactivity. 

since app router defaults to rendering components as server components, I think John's suggestion below about moving the Map into its own component file (with the "use client" directive at the top of that file) would allow removing the directive at the top of this file without that error

signs-work-order-map/app/page.tsx
mddilley: that is a good idea! i often forget that we have access to browserstack. I think that the new map should have the fancy implementation and could be a future issue. since [the original behaves that way](https://atd-knack-signs-markings.netlify.app/), i'm hoping that is a sign that it won't be a performance problem. 🤞

signs-work-order-map/env_template
mddilley: i think the VZE consitency point is a good one. i think Next will use whichever one it finds so i think it is a matter of just renaming the file

mddilley (commented):
Rad to have this map to start this rebuild 💪 And to have the geocoder in here already is so nice! I left some comments - this was a fun one to read and think through. I think this is a great start as a collector branch. 🗺️

signs-work-order-map/app/page.tsx
mateoclarke: I was expecting the pin to stay centered like I think John is describing with `onDrag`. It's really hard to guess where you're pin will land without that real-time feedback. And the way the pin jumps around is confusing. But maybe all this was going to be addressed in a follow on feature.

signs-work-order-map/components/MapGeocoderControl.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+  const geocoder = useControl<MapboxGeocoder>(
+    () => {
+      const ctrl = new MapboxGeocoder({
+        ...props,
+        marker: false,
+        accessToken: MAPBOX_TOKEN || "",
+        bbox: ATX_BOUNDING_BOX,
+      });
+      // ctrl.on('loading', props.onLoading);
```
mateoclarke: maybe this was just copy/pasta but will we need these commented lines?

signs-work-order-map/components/MapGeocoderControl.tsx
```diff
@@ -0,0 +1,146 @@
...
+      // ctrl.on('results', props.onResults);
+      ctrl.on("result", (evt) => {
```
mateoclarke: I noticed above we call this param `e`, here we call it `evt`. My preference would be to consistency and verbosity with `event`

signs-work-order-map/package.json
```diff
@@ -0,0 +1,31 @@
+{
+  "name": "signs-work-order-map",
+  "version": "0.1.0",
+  "private": true,
+  "scripts": {
+    "dev": "next dev",
```
mateoclarke: How would we feel about picking a non-default port number for this to run locally that doesn't conflict with VZE (`:3002`) or Moped Editor (`:3000`)?

For example:
`"dev": "next dev -p 3003",`

mateoclarke (commented):
This looks good to me as a starting place/collector branch. 👍 

paul-ryan-good-start

signs-work-order-map/app/page.module.css
johnclary: Yeah, it'd be nice to clean up the CSS so that we have a clean starting point. 

Fwiw, bootstrap is coming into VZE at ~26kb, which is roughly the same size as a single nearmap tile.

signs-work-order-map/package.json
johnclary: yeah! thank you Chia for sorting out our upgrade path to v8 in the VZE 😇

signs-work-order-map/.eslintrc.json
chiaberry: I added the eslint plugins like vz. now im getting an error message when i run `npm run lint`
`ESLint configuration in ../package.json » eslint-config-react-app is invalid:`, I'll see where the mismatch is.

signs-work-order-map/package.json
chiaberry: I tried updating the package.json with this, but then I got 403s with the mapbox tiles. I dont understand why....

signs-work-order-map/app/page.module.css
chiaberry: I cleaned up this file, but have not included bootstrap yet

signs-work-order-map/app/page.tsx
chiaberry: I switched it to be onDrag, eventually I am thinking the address lookup will happen onMoveEnd, but that's a future question for future devs

signs-work-order-map/package.json
johnclary: I wonder if there's an IP whitelist for mapbox (i'm surprised that 3001 works with the VZE, though). there definitely is for Nearmap when it comes time to get that key set up

signs-work-order-map/app/page.tsx
johnclary: The drag effect looks great 🙏 

This can be a future enhancement—and VZE has the same issue—it would good to keep the pin centered when zooming in/out as well. Currently the pin location doesn't update when you zoom, and then the next time you pan the map the pin jumps to center.

johnclary:
Updates look good! Should we hold off on approving, since this is a collector? Or can `staging` be the collector branch (since there's no CI?)?

chiaberry:
@johnclary I was thinking of not merging this branch, but you are right since there is no CI, we could use staging as the collector branch! Now I am not sure.

ChristinaTremel:
> Something I completely overlooked when reviewing this repo last month was the [calcs file](https://github.com/cityofaustin/atd-knack-signs-markings/blob/staging/src/calcs.js), which renders some calculation widgets https://atd-knack-signs-markings.netlify.app/calcs, embedded in a page in knack. Christina is checking if this is still being used by SMD and if it needs to be included when we redeploy.

I followed up with the SMD Supervisors and they confirmed that technicians are in fact still using this markings material calculator. 🙃

signs-work-order-map/.gitignore
frankhereford, line 1: What would you think about paring down this `.gitignore` to only include the things that we're actively ignoring with it. These "catch all" type ignore files are a lot more noise than signal to me, and they are inevitably incomplete.   Seeing `vercel` and `yarn` in here prompted the thought -- those are technologies we don't use on this team to my knowledge.

signs-work-order-map/config/map.ts
```diff
@@ -0,0 +1,127 @@
+import "mapbox-gl/dist/mapbox-gl.css";
+// import { SymbolLayerSpecification, RasterLayerSpecification } from "mapbox-gl";
+
+// The Nearmap API key is managed by CTM. Contact help desk for maintenance and troubleshooting.
+const NEARMAP_KEY = process.env.NEXT_PUBLIC_NEARMAP_KEY;
```
frankhereford: You may still plan on incorporating it, and if so, please ignore this, but this key doesn't look like it's used anywhere.

signs-work-order-map/env_template
frankhereford, line 1: EOF linefeed is missing here

signs-work-order-map/package.json
frankhereford: Traditionally, these slippy-map keys have been restricted by what the referrer sent by the client when requesting resources is. To me, it's hand-wavy security, because it's pretty easy to write a proxy for them that you can use to mask the actual referrer sent from the browser, or to even just redefine the browser's own referrer behavior. That said, I can see that if you don't want to wade into the complexity of JWTs or similar, you gotta just ship that API key out to the client and hope for the best.

Anyway, Mapbox tokens are not required, but may include, these referrer restrictions. See the following screenshot from the token issuance page for some more info.

Screenshot 2025-04-21 at 11 26 33 AM

frankhereford (commented):
This is looking awesome Chia, and I can't wait to see what is next. I know you've already got a ton of work done past this point. Thank you! 

To answer your question, I think this is a great looking collector branch.

PR #319, opened 2022-11-01
Remove circleci and update readme

chiaberry:
As part of accomplishing https://github.com/cityofaustin/atd-data-tech/issues/8788, I have deprecated the circleCI integration. 

Due to the lack of active development on this repo I've chosen to not create a github action to push changes to s3. Changes should be manually uploaded.

PR #326, opened 2025-05-07
iFrame message communication

chiaberry:
Related issue: https://github.com/cityofaustin/atd-data-tech/issues/22245

This PR sets up the communication between the signs and markings Knack application and our Next JS app. I have refactored where the knack record fetching is happening. Currently in production, custom javascript in knack loads the iFrameMessenger.js script (stored in s3), which sends payloads from knack to the react app. The react app then [requests data from knack](https://github.com/cityofaustin/atd-knack-signs-markings/blob/staging/src/SelectLocation.js#L401) depending on what kind of message is being sent. Depending on the view that contains the react app, the app requests signs from work order records and/or location information. 

I refactored it so the knack requests happen in the iFrameMessenger script and the payload includes the array of signs and the location data. 

### Testing instructions

1. sign in as a non-coa user, `test-tech@austintexas.gov`, the password is in 1pw under SMD test technician
2. Open the dev tools console, messages received from the knack are currently only being console.logged in the react app. main.js is the knack app, iframeMess...is the script in s3 and page-blahblah is the react app.
3. If you have location services turned on, and the site has permission to get your location, you should see the KNACK_GEOLOCATION message. You will also see "autozoom message to app  KNACK_GEOLOCATION" but that is a lie, nothing is autozooming yet. 
4. Find a work order with signs, for example: https://atd.knack.com/test-30-may-2024-signs-and-markings-operations#work-order-signs/view-work-orders-details-sign/661efebdd942c30029c40014/ . Check that you receive an array of sign records. (Eventually these signs will be rendered on the map) This is **view_2573**. In production this payload is called signs_api_request, but I have renamed it as WORK_ORDER_SIGNS
5. Click into one of the location detail pages. ([example from above](https://atd.knack.com/test-30-may-2024-signs-and-markings-operations#work-order-signs/view-work-orders-details-sign/661efebdd942c30029c40014/view-work-order-signs-location-details/6645069c20daa1002997b71b/)). 

Screenshot 2025-05-28 at 11 05 05 AM

This is **view_2733** which is called KNACK_LOCATION_DETAILS in prod and I have kept the same name. The payload should be all the signs from before but also the location's lat/lon. If you zoom out, the location should be shown on this map. (Eventually, the map will center on the location)

6. At the top of this location detail page, there is a small edit link. Click that to test **view_2682**. In prod this is called EDIT_SIGNS_API_REQUEST, but I've renamed it to EDIT_LOCATION. The payload is only the location information. Again if you zoom out, the location is on the map. If you drag the map around, the default marker's location updates the location fields, information is coming from the react app and going to knack. 

7. You can also test communication from react to knack by going to a work order (example:https://atd.knack.com/test-30-may-2024-signs-and-markings-operations#work-order-signs/view-work-orders-details-sign/6655b8e50480d80276a8bb45/) and moving the map around, and then clicking "Add location" 
Screenshot 2025-05-28 at 11 20 17 AM. 
Knack will add a new location based on where your map is centered. 

If you want to run the react app locally and test that way, you need to update the iFrameMapMessengerDev.js file in s3 to point [to localhost as the iframe ](https://github.com/cityofaustin/atd-knack-signs-markings/pull/326/files#diff-770d4a37ba11c473af32738210b093c1664103576fbc1f38e703b4c33b6f1e27R48). Make sure it is the dev file and not the production file.


knack/iframeMapMessengerDev.js
```diff
@@ -0,0 +1,162 @@
...
+          message: "KNACK_LOCATION_DETAILS",
+          view: "view_2733",
+          scene: "scene_1039",
+          token: Knack.getUserToken(),
+          app_id: Knack.application_id,
+          id: recordId,
+          workOrderScene: "scene_1028",
+          workOrderId: workOrderId,
+          workOrderView: "view_2573"
+        };
```
chiaberry: I am thinking about moving some of this payload into the react app side, if the view and scenes are hard coded, why not store them in code that is easier to update

knack/iframeMapMessengerDev.js
chiaberry, line 1: this is mostly the same as what exists in production now, except for some comments and this line that quiets the incessant console logging that happens in prod right now
```
        if (event.data.source === "react-devtools-content-script") {
          return;
        }
```

chiaberry: completely no longer true

knack/iframeMapMessenger.js
```diff
@@ -64,61 +84,66 @@
...
-        view: "view_2733",
-        scene: "scene_1039",
-        token: Knack.getUserToken(),
-        app_id: Knack.application_id,
-        id: recordId,
-        workOrderScene: "scene_1028",
-        workOrderId: workOrderId,
-        workOrderView: "view_2573"
+        records: [],
+        location: [],
```
chiaberry: I am wondering if instead this should have the format of 
```javascript
{
  message: "KNACK_LOCATION_DETAILS",
  payload: {
     records : [],
     location: []
  }
}
```
because as I am looking at the issue to render signs on the map, I am finding we may need more information like work Order ID, or knack view

chiaberry: I found this https://www.nickwhite.cc/blog/strongly-typed-iframe-messaging/, which is what is nudging me to make the types better

knack/README.md
```diff
@@ -10,4 +10,4 @@ November 1, 2022 deprecated CircleCI integration and removed [config files](http
 
 Any updates to the iframeMapMessenger.js file should be manually uploaded to the s3 bucket `atd-knack-signs-markings` in the DTS AWS account.
 
-https://atd-knack-signs-markings.s3.amazonaws.com/staging/iframeMapMessenger.js
\ No newline at end of file
+https://atd-knack-signs-markings.s3.amazonaws.com/staging/iframeMapMessenger.js
```
johnclary, line 13: i guess we never created a `/prod` version of this script?

knack/iframeMapMessenger.js
```diff
@@ -10,46 +10,66 @@
...
 
     // Add React app as iframe if iframe doesn't already exist
     if ($(myView + " #mapIFrame").length === 0) {
       https: $(
-        '<iframe src="https://atd-knack-signs-markings.netlify.app/" frameborder="0" scrolling="yes" id="mapIFrame" \
-    style="width: 100%;height: 523px;"></iframe>'
+        // '<iframe src="https://atd-knack-signs-markings.netlify.app/" frameborder="0" scrolling="yes" id="mapIFrame" \
+        // '<iframe src="https://deploy-preview-326--nextjs-knack-signs-markings.netlify.app/" frameborder="0" scrolling="yes" id="mapIFrame" \
+        '<iframe src="http://localhost:3000/" frameborder="0" scrolling="yes" id="mapIFrame" \
+      style="width: 100%;height: 523px;"></iframe>'
```
johnclary: Do we know yet what device make/model the technicians are using?

knack/iframeMapMessenger.js
```diff
@@ -10,46 +10,66 @@
...
-      console.log("inside API", stringifiedMessage);
       iframe.postMessage(stringifiedMessage, "*");
     }
 
     // Listen for lat/lon changes
-    window.addEventListener("message", function(event) {
-      console.log("message received:  " + event.data, event);
+    // expects a message named "LAT_LON_FIELDS"
+    // uses lat and lng from message to populate input fields in knack
+    window.addEventListener("message", function (event) {
```
johnclary, line 73: as a security measure, i believe it is customary to validate the event `origin` before handling a message. something like: 

```javascript
if (event.origin !== 'https://myreact-app.com') { return; }
```

That might rule out the need to handle the react-devtools source down below 🤷

knack/iframeMapMessenger.js
johnclary: I like that pattern!

knack/iframeMapMessenger.js
```diff
@@ -64,61 +84,66 @@
...
-        workOrderView: "view_2573"
+        records: [],
+        location: [],
       };
 
-      sendMessageToApp(markerMessage, locationViewIFrame);
+      // Request the location based on record ID
+      console.log("Requesting records for location id ", recordId);
+      $.ajax({
+        url: `https://api.knack.com/v1/scenes/scene_1039/views/view_2733/records/${recordId}`,
```
johnclary, line 106: it might be nice to stash these magic scene and view numbers in vars outside the main function. Ditto for the other scene, view, and field IDs in this code.

knack/iframeMapMessenger.js
```diff
@@ -64,61 +84,66 @@
...
+          locationField.latitude,
+          locationField.longitude,
+        ];
+        console.log("requesting records for work order id ", workOrderId);
+        // Request the associated signs records
+        $.ajax({
+          url: `https://api.knack.com/v1/scenes/scene_1028/views/view_2573/records?view-work-orders-details-sign_id=${workOrderId}`,
+          headers: headers,
+        }).then(function (res) {
+          var records = res.records;
```
johnclary: this var seems redundant versus assigning directly to signsMarkerMessage.records

knack/iframeMapMessenger.js
johnclary, line 1: it might be nice to add a note at the top of this that explains that this code is imported into the Knack app and executed as custom vanilla JS that interacts with the NextJS app via iFrame messaging.

this would not be a trivial lift, but since we import this script from S3 we could convert it to typescript and compile it as javscript during our build process 🤔

knack/iframeMapMessenger.js
```diff
@@ -64,61 +84,66 @@
...
+        headers: headers,
+      }).then(function (res) {
+        var locationField = res["field_3300_raw"];
+        signsMarkerMessage.location = [
+          locationField.latitude,
+          locationField.longitude,
+        ];
+        console.log("requesting records for work order id ", workOrderId);
+        // Request the associated signs records
+        $.ajax({
```
johnclary: not a big deal, but i was expecting to see this block return the ajax promise so that the next `then` can be linked to a single chain of promises (instead of nested promises). this would also let you add a single `catch` handler for all the requests—which might be good to add?

knack/iframeMapMessenger.js
```diff
@@ -128,28 +153,58 @@
...
+      }).then(function (res) {
+        var locationField = res["field_3300_raw"];
+        console.log("EDIT_LOCATION: ", locationField);
+        var locationMessage = {
+          message: "EDIT_LOCATION",
+          records: [],
+          location: [locationField.latitude, locationField.longitude],
+        };
+        sendMessageToApp(locationMessage, editLocationIframe);
+      });
```
johnclary: should we add error handling for all these ajax requests? i wonder how often these Knack API calls fail...

signs-work-order-map/utils/iFrameMessenger.tsx
```diff
@@ -0,0 +1,33 @@
...
+  records: [];
+  location: [number, number];
+};
+
+export function useIFrameMessenger() {
+  const [message, setMessage] = useState<MessageToIframe | null>(null);
+
+  useEffect(() => {
+    const consoleMessage = (event: MessageEvent) => {
+      if (
```
johnclary, line 15: same here—i think it is customary to check the event `origin`

knack/iframeMapMessenger.js
```diff
@@ -64,61 +84,66 @@
...
+      // Request the location based on record ID
+      console.log("Requesting records for location id ", recordId);
+      $.ajax({
+        url: `https://api.knack.com/v1/scenes/scene_1039/views/view_2733/records/${recordId}`,
+        headers: headers,
+      }).then(function (res) {
+        var locationField = res["field_3300_raw"];
+        signsMarkerMessage.location = [
+          locationField.latitude,
+          locationField.longitude,
```
johnclary: are we sure `locationField` field is never null? In other words, can this `signsMarkerMessage.location` ever be assigned as `[undefined, undefined`]?

signs-work-order-map/components/Map.tsx
```diff
@@ -14,22 +14,36 @@ interface LatLon {
   longitude: number;
 }
 
-export default function Map() {
+interface MapProps {
+  location: [number, number] | undefined;
```
johnclary, line 18: just want to double check that we are sure we will never have to handle `[undefined, undefined]`

signs-work-order-map/utils/iFrameMessenger.tsx
```diff
@@ -0,0 +1,33 @@
...
+        event.data.source === "react-devtools-content-script" ||
+        event.data.source === "react-devtools-bridge" ||
+        event.data.source === "react-devtools-backend-manager"
+      ) {
+        return;
+      }
+      const data = JSON.parse(event.data);
```
johnclary, line 22: i'm not sure how much we want to concern ourselves with this, but data is typed as `any` here and we are assuming that it conforms to the `MessageToIframe` type. we could add type guards here if we wanted to 🤷

knack/iframeMapMessenger.js
```diff
@@ -64,61 +84,66 @@
...
+      var signsMarkerMessage = {
         message: "KNACK_LOCATION_DETAILS",
-        view: "view_2733",
-        scene: "scene_1039",
-        token: Knack.getUserToken(),
-        app_id: Knack.application_id,
-        id: recordId,
-        workOrderScene: "scene_1028",
-        workOrderId: workOrderId,
-        workOrderView: "view_2573"
```
johnclary, line 76: i really like that you moved the data fetching elsewhere and simplified the message payload 👍

johnclary (approved):
This is looking great! It's very fun to see the messages flowing for real now ✨ 

I'd love to know what device the techs are using so we can be thinking about usability while we're testing this. A lot of the buttons/links are pretty small.

I also wonder if we might want to turn on the Mapbox feature that prevents scroll capture. I noticed i kept zooming the map when I tried to scroll the work order details page. Here's what this looks like in VZE on a mac:

scroll-saver

knack/iframeMapMessenger.js
chiaberry: great question -- i am going to check. 

also I need to remember to flip this to  return [long, lat]

knack/iframeMapMessenger.js
mddilley: +1 - i love this change! i think it makes it clear separation between the message or action and its associated data - whatever that might be. 🚀

knack/iframeMapMessenger.js
mddilley: +1 and this could make it easier to move this code between environments too if we move this code out of S3 in the future.

knack/README.md
mddilley: yeah, we did not. This is also making me wonder if we  should continue to store this code in S3 or if the code could live in the Knack app custom code. 

iirc this file was intended to make development a lower lift but 200 lines of JS doesn't seem too extreme now that we have some years of custom Knack work under our belts. 🤷

definitely a future issue and i might be missing an important detail about S3! it has been a while 😅

knack/iframeMapMessenger.js
mddilley: yep, agreed here, we do this on the React side (see below) in the current prod app but we missed the detail of checking on the other Knack end.

```js
if (event.origin !== "https://atd.knack.com") return;
```

This would probably be another good variable to live outside the JS to help us test on deploy previews.

edit: I think this line would still work well for John's suggestion about adding the origin check on the React side!

knack/iframeMapMessenger.js
```diff
@@ -64,61 +84,66 @@
...
+      }).then(function (res) {
+        var records = res.records;
+        console.log("WORK ORDER SIGNS: ", records);
+        var signsMarkerMessage = {
+          message: "WORK_ORDER_SIGNS",
+          records: records,
+          location: [],
+        };
+        sendMessageToApp(signsMarkerMessage, workOrderDetailsIFrame);
+      });
```
mddilley: yes! I think that Knack data fetching in the Knack code will make this much easier to understand for the next person instead of the payload being the credentials needed to make a request for that data. 🙏

signs-work-order-map/components/Map.tsx
```diff
@@ -14,22 +14,36 @@ interface LatLon {
...
+    const longitude = +event.viewState.longitude.toFixed(
+      MAP_COORDINATE_PRECISION
+    );
     console.log(latitude, longitude);
     setMapLatLon({
       latitude,
       longitude,
     });
+    // send location to Knack
+    window.parent.postMessage(
```
mddilley, line 36: part of me thinks that wrapping this postMessage function as a function that is returned from the `useIFrameMessenger` custom hook might help us complete the loop on sending and receiving messages but idk. 

I also go wild with grouping things with custom hooks sometimes... 🙃

mddilley (approved):
So fun to see the next-gen iframe messenger. 😎 Love that it turned into a custom hook 🪝 All the test steps check out, and this is a really nice improvement. 🙌 I added a few comments and suggestions.

knack/iframeMapMessenger.js
ChristinaTremel: They are using iPad minis and I _think_ they are using 6th generation models!

roseeichelmann (approved):
test steps checked out, this has been a lott of work, hugee props to you for figuring it all out!! fwiw to address johns comment about the scrolling, i also tested this on my phone since thats more similar to an ipad and that wasnt an issue, i think it would only be an issue on desktops

frankhereford (approved):
Wow, this is really great. This is my first time in the app proper and seeing all the functionality up close and personal. 

Your testing steps were clear, and I was able to work through them with no issue at all. 

I've gone through the code, and I wish I had something to offer, but I'm rusty and it looked good to me. Previous reviewers have left good feedback.

I can't wait to see where this goes next!

knack/README.md
mateoclarke: Wasn't this automated at one point? So whatever was in `main` would get pushed to S3?

mateoclarke (approved):
Looks great Chia! Thanks for bringing this app into modernity, lol. 🚀

knack/README.md
mddilley: oh yeah, this jogged my memory! 🙏 We didn't migrate the CI to GitHub Actions when we moved fully away from CircleCI across our repos - https://github.com/cityofaustin/atd-knack-signs-markings/pull/319.

PR #327, opened 2025-06-04
Render signs on map 

chiaberry:
Related Issue: https://github.com/cityofaustin/atd-data-tech/issues/22803

Testing instructions

Work orders have sign records if they have the status "Final Review" or "Closed"
Screenshot 2025-06-13 at 2 00 27 PM

Click on a work order, when the page loads, the map should zoom to show the sign markers. A lot of the records only have one location, but [here](https://atd.knack.com/test-30-may-2024-signs-and-markings-operations#work-order-signs/view-work-orders-details-sign/5fa5b8264416d10019334024/) is an example of a work order with multiple locations. 

Clicking on one of the markers will bring up the details popup, with a link to the location details page. Click the link to the details page. 

Screenshot 2025-06-13 at 2 11 38 PM

The map should look very similar on the location details page, except for the marker color. The location that corresponds to the location detail page being viewed should have a red marker, and the popup omits the link to the location details page. 

What do you think of the popup's styling? Production has everything bolded on it (see issue for reference), do you think it should be bolder? Font bigger?

Like the previous PR, if you want to run the react app locally and test that way, you need to update the iFrameMapMessengerDev.js file in s3 to use localhost instead of the deployed site. 
Make sure it is the dev file and not the production file.

### DIFF
--- a/knack/iframeMapMessenger.js
+++ b/knack/iframeMapMessenger.js
@@ -114,6 +114,7 @@
           signsMarkerMessage.payload.location.latitude = locationField.latitude;
           signsMarkerMessage.payload.location.longitude =
             locationField.longitude;
+          signsMarkerMessage.payload.locationRecordId = recordId;
         })
         .then(function () {
           console.log("requesting records for work order id ", workOrderId);
--- a/knack/iframeMapMessenger.js
+++ b/knack/iframeMapMessenger.js
@@ -7,8 +7,9 @@
 (function () {
   var myView = window.viewIdsArray.shift(0);
 
-  // const nextAppUrl = "https://deploy-preview-326--nextjs-knack-signs-markings.netlify.app/";
-  const nextAppUrl = "http://localhost:3000";
+  const nextAppUrl =
+    "https://deploy-preview-327--nextjs-knack-signs-markings.netlify.app/";
+  // const nextAppUrl = "http://localhost:3000";
 
   // Import jQuery into this file from CDN
   // https://stackoverflow.com/questions/34338411/how-to-import-jquery-using-es6-syntax
@@ -111,7 +112,8 @@
       })
         .then(function (res) {
           var locationField = res["field_3300_raw"];
-          signsMarkerMessage.payload.location.latitude = locationField?.latitude;
+          signsMarkerMessage.payload.location.latitude =
+            locationField?.latitude;
           signsMarkerMessage.payload.location.longitude =
             locationField?.longitude;
           signsMarkerMessage.payload.locationRecordId = recordId;
--- a/knack/iframeMapMessenger.js
+++ b/knack/iframeMapMessenger.js
@@ -96,7 +96,10 @@
         message: "KNACK_LOCATION_DETAILS",
         payload: {
           records: [],
-          location: [],
+          location: {
+            longitude: undefined,
+            latitude: undefined,
+          },
         },
       };
 
@@ -107,11 +110,11 @@
         headers: headers,
       })
         .then(function (res) {
-          var locationField = res["field_3300_raw"]; // check if this is undefined
-          signsMarkerMessage.location = [
-            locationField.latitude,
-            locationField.longitude,
-          ];
+          var locationField = res["field_3300_raw"];
+          signsMarkerMessage.payload.location.longitude =
+            locationField.latitude;
+          signsMarkerMessage.payload.location.longitude =
+            locationField.longitude;
         })
         .then(function () {
           console.log("requesting records for work order id ", workOrderId);
@@ -121,7 +124,7 @@
             headers: headers,
           }).then(function (res) {
             var records = res.records;
-            signsMarkerMessage.records = records;
+            signsMarkerMessage.payload.records = records;
             sendMessageToApp(signsMarkerMessage, locationViewIFrame);
           });
         })
@@ -153,7 +156,6 @@
             message: "WORK_ORDER_SIGNS",
             payload: {
               records: res.records,
-              location: [],
               workOrderId: recordId,
             },
           };
@@ -186,8 +188,10 @@
           var locationMessage = {
             message: "EDIT_LOCATION",
             payload: {
-              records: [],
-              location: [locationField.latitude, locationField.longitude],
+              location: {
+                longitude: locationField.longitude,
+                latitude: locationField.latitude,
+              },
             },
           };
           sendMessageToApp(locationMessage, editLocationIframe);
@@ -227,9 +231,10 @@
       const geolocationMessage = {
         message: "KNACK_GEOLOCATION",
         payload: {
-          records: [],
-          location: [],
-          geolocation: [position.coords.latitude, position.coords.longitude],
+          geolocation: {
+            latitude: position.coords.latitude,
+            longitude: position.coords.longitude,
+          },
         },
       };
 
--- a/knack/iframeMapMessenger.js
+++ b/knack/iframeMapMessenger.js
@@ -111,8 +111,7 @@
       })
         .then(function (res) {
           var locationField = res["field_3300_raw"];
-          signsMarkerMessage.payload.location.longitude =
-            locationField.latitude;
+          signsMarkerMessage.payload.location.latitude = locationField.latitude;
           signsMarkerMessage.payload.location.longitude =
             locationField.longitude;
         })
@@ -125,6 +124,7 @@
           }).then(function (res) {
             var records = res.records;
             signsMarkerMessage.payload.records = records;
+            signsMarkerMessage.payload.workOrderId = workOrderId;
             sendMessageToApp(signsMarkerMessage, locationViewIFrame);
           });
         })
--- a/knack/iframeMapMessenger.js
+++ b/knack/iframeMapMessenger.js
@@ -111,9 +111,9 @@
       })
         .then(function (res) {
           var locationField = res["field_3300_raw"];
-          signsMarkerMessage.payload.location.latitude = locationField.latitude;
+          signsMarkerMessage.payload.location.latitude = locationField?.latitude;
           signsMarkerMessage.payload.location.longitude =
-            locationField.longitude;
+            locationField?.longitude;
           signsMarkerMessage.payload.locationRecordId = recordId;
         })
         .then(function () {
@@ -190,8 +190,8 @@
             message: "EDIT_LOCATION",
             payload: {
               location: {
-                longitude: locationField.longitude,
-                latitude: locationField.latitude,
+                longitude: locationField?.longitude,
+                latitude: locationField?.latitude,
               },
             },
           };
--- a/signs-work-order-map/app/page.tsx
+++ b/signs-work-order-map/app/page.tsx
@@ -1,73 +1,31 @@
 "use client";
+import { useMemo } from "react";
 import styles from "./page.module.css";
 import Map from "@/components/Map";
 import { KnackRecord, Sign } from "@/types/map";
 import { useIFrameMessenger } from "@/utils/iFrameMessenger";
 
-const testpayload = {
-  location: undefined,
-  message: "WORK_ORDER_SIGNS",
-  records: [
-    {
-      field_3297: 202,
-      field_3297_raw: 202,
-      field_3300: "4702 Oldfort Hill Dr<br />Austin, Texas 78723",
-      field_3300_raw: {
-        city: "Austin",
-        country: "United States",
-        full: "4702 Oldfort Hill Dr Austin, Texas 78723",
-        latitude: 30.286771,
-        longitude: -97.675177,
-        state: "Texas",
-        street: "4702 Oldfort Hill Dr",
-        street2: null,
-        zip: "78723",
-      },
-      field_3301:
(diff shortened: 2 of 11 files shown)


knack/iframeMapMessenger.js
```diff
@@ -7,8 +7,9 @@
 (function () {
   var myView = window.viewIdsArray.shift(0);
 
-  // const nextAppUrl = "https://deploy-preview-326--nextjs-knack-signs-markings.netlify.app/";
-  const nextAppUrl = "http://localhost:3000";
+  const nextAppUrl =
+    "https://deploy-preview-327--nextjs-knack-signs-markings.netlify.app/";
+  // const nextAppUrl = "http://localhost:3000";
```
chiaberry, line 12: here is where you can switch if you want to run locally

signs-work-order-map/config/map.ts
```diff
@@ -24,6 +24,7 @@ export const DEFAULT_MAP_PARAMS = {
   boxZoom: false,
   mapboxAccessToken: MAPBOX_TOKEN,
   maxBounds: MAP_MAX_BOUNDS,
+  maxZoom: 20,
```
chiaberry: it zooms in very close when there is only one sign, so I added this. open to feedback!

signs-work-order-map/components/SignPopup.tsx
```diff
@@ -0,0 +1,40 @@
+import { Popup } from "react-map-gl/mapbox";
+import { Sign } from "@/types/map";
+
+interface SignPopupProps {
+  popupInfo: Sign;
+  setPopupInfo: React.Dispatch<React.SetStateAction<Sign | null>>;
+}
+
+export default function SignPopup({ popupInfo, setPopupInfo }: SignPopupProps) {
+  // prod has the font as bold, do we keep?
```
roseeichelmann, line 11: i prefer the text not being bold! i think yours looks better than prod

signs-work-order-map/config/map.ts
roseeichelmann: i feel like it still zooms in pretty close!! could be zoomed out even more imo

roseeichelmann (approved):
followed the test steps and this looks good!

is it in scope to add the zoom in/out buttons to the map, or is that gonna be in a diff issue? also, it throws me off that the map always starts zoomed in to that spot on 15th street, and then sometimes it takes a long time to zoom to the actual sign marker. but idk if thats intentional for now and you plan on addressing that later?

signs-work-order-map/components/Map.tsx
```diff
@@ -44,24 +42,53 @@ export default function Map({ location }: MapProps) {
...
+
+  const signPins = createSignPins(signs, setPopupInfo);
+
+  useEffect(() => {
+    if (!mapRef?.current || !bounds) {
+      return;
+    }
+    mapRef.current.fitBounds(bounds, {
+      padding: 100,
+    });
```
mddilley: I remember in Moped we decided to override the animation when setting the initial bounds since it could get jarring when moving from project to project. I think that the animation duration can be set to 0 in these options iirc.

signs-work-order-map/components/SignPopup.tsx
mddilley: +1 about the bold. I think that the font could be bigger but maybe that is a future convo about responsiveness & screen sizes.

signs-work-order-map/config/map.ts
mddilley: another option is to set this in the options for `fitBounds` if you want to control the zoom on that level only ([docs](https://docs.mapbox.com/mapbox-gl-js/api/map/#map%23fitbounds)) .  Having here could have other benefits that I'm missing, though.

signs-work-order-map/components/SignPopup.tsx
```diff
@@ -0,0 +1,40 @@
...
+      longitude={popupInfo.lng}
+      latitude={popupInfo.lat}
+      onClose={() => setPopupInfo(null)}
+      offset={[0, -20]}
+    >
+      <ul className="list-unstyled m-0">
+        {!popupInfo.locationDetailPage && (
+          <li>
+            <a
+            // TODO: update to prod url when testing is completed
```
mddilley: maybe it would be beneficial to set this url in a variable somewhere? idk how many times the Knack app link is used around the app but could help with future development

signs-work-order-map/types/map.ts
```diff
@@ -0,0 +1,107 @@
...
+  lng: number;
+  lat: number;
+  spatialId: number;
+  workOrderId: string;
+  locationDetailPage: boolean;
+}
+
+export interface KnackRecord {
+  /** spatial id */
+  field_3297: number;
```
mddilley, line 25: I like how this serves as documentation of the Knack fields! Is there any possibility of these field numbers changing between prod and the test apps? It has been way too long since I've been in the Knack world. 😅

signs-work-order-map/utils/mapUtils.tsx
```diff
@@ -0,0 +1,121 @@
...
+import { LngLatBoundsLike } from "mapbox-gl";
+import { Marker } from "react-map-gl/mapbox";
+import { Sign, KnackToIFrameMessage, LatLon } from "@/types/map";
+
+/**
+ * Takes array of Signs and if signs exist, returns bounding box for signs
+ * @param signs
+ * @returns bbox extent in [minX, minY, maxX, maxY] order or undefined
+ */
+export const formatBounds = (signs: Sign[]): undefined | LngLatBoundsLike =>
```
mddilley: It might be nice to add `use` to the start of the custom hook functions in this file. I really love seeing how all this data is built up btw! 🚀

signs-work-order-map/utils/mapUtils.tsx
```diff
@@ -0,0 +1,121 @@
...
+        ? knackPayload?.payload?.locationRecordId
+        : null;
+
+    return knackPayload.payload.records.map((sign) => ({
+      id: sign.id,
+      lat: sign.field_3300_raw.latitude,
+      lng: sign.field_3300_raw.longitude,
+      spatialId: sign.field_3297,
+      workOrderId: knackPayload.payload.workOrderId,
+      locationDetailPage: sign.id === locationId,
```
mddilley: I think adding `is` to the beginning of `locationDetailPage` could make it clearer that it is a boolean being checked in the `createSignPins` hook. 

This is a really nifty way to control the red or blue color on the pins!! ✨

mddilley (approved):
This works great! Really cool to see more views where this code is used coming together. I left a few comments and questions, but nothing holding this back. Very cool stuff! 😎

signs-work-order-map/types/map.ts
chiaberry: good point, field numbers could stray if fields are being added to prod and then different ones are added to test before test is brought up to match prod. I don't think knack development is happening, but its something to be aware of

signs-work-order-map/utils/mapUtils.tsx
chiaberry: ahhh yes I am so bad at remembering to name these correctly

johnclary: +1 — definitely needed. the hooks look great!

- `formatBounds`
- `formatSignsRecords`
- `formatLocation`
- `createSignPins`

signs-work-order-map/components/Map.tsx
```diff
@@ -44,24 +42,53 @@ export default function Map({ location }: MapProps) {
...
+
   return (
     <MapGL
+      ref={mapRef}
       initialViewState={{
         latitude: DEFAULT_MAP_PAN_ZOOM.latitude,
         longitude: DEFAULT_MAP_PAN_ZOOM.longitude,
         zoom: DEFAULT_MAP_PAN_ZOOM.zoom,
       }}
+      cooperativeGestures={true}
```
johnclary, line 69: 🖱️

signs-work-order-map/app/page.tsx
```diff
@@ -2,19 +2,21 @@
...
-  const location = knackPayload?.location;
+  const location = useFormatLocation(knackPayload);
+  const signs = useFormatSignsRecords(knackPayload);
 
   return (
     <div className={styles.page}>
       <main className={styles.main}>
         <div className={styles.map}>
-          <Map location={location} />
+          <Map location={location} signs={signs} />
```
johnclary, line 19: I got a little confused about the nomenclature, because it seems that the Knack app calls these "locations" instead of signs, and a location can have multiple sign assets (ie the location is the sign pole). Is it worth aligning this terminology in the nextjs app?

And I guess `location` here is always the user's location?

signs-work-order-map/components/SignPopup.tsx
```diff
@@ -0,0 +1,40 @@
...
+              target="_top"
+            >
+              Location Detail Page
+            </a>
+          </li>
+        )}
+        <li>Spatial ID: {popupInfo.spatialId}</li>
+        <li>Latitude: {popupInfo.lat}</li>
+        <li>Longitude: {popupInfo.lng}</li>
+      </ul>
```
johnclary, line 37: I know it's out of scope, but fwiw these pop-ups might have some usability issues due to how small they are.

Screenshot 2025-06-18 at 2 38 08 PM

If it's helpful, on https://data.mobility.austin.gov/ we use modal for popups on smaller displays

Screenshot 2025-06-18 at 3 02 32 PM

Screenshot 2025-06-18 at 3 02 27 PM

signs-work-order-map/config/map.ts
johnclary: Hmm yeah the initial map zoom on one sign is real close in. I assume technicians do need the ability to zoom way far in order to place sign locations correctly, or to click on the right location when they are very close together.

signs-work-order-map/types/map.ts
```diff
@@ -0,0 +1,107 @@
...
+    longitude: number | undefined;
+    latitude: number | undefined;
+  };
+  signs: Sign[];
+}
+
+export interface Sign {
+  id: string;
+  lng: number;
+  lat: number;
```
johnclary, line 17: Chia and I discussed an edge case error I ran into when I clicked on the **Add Location** button (even though feature is out of scope). I wonder if these lat/lon need to be optional?

signs-work-order-map/types/map.ts
```diff
@@ -0,0 +1,107 @@
...
+  field_3378: string;
+  /** photo raw format */
+  field_3378_raw: string;
+  /** count of assets  */
+  field_3425: number;
+  /** count of assets raw format*/
+  field_3425_raw: number;
+  /** knack record id */
+  id: string;
+}
```
johnclary, line 68: Are we sure that all of these fields will all be defined? To err on the side of caution you could make them all optional (`?`)

signs-work-order-map/types/map.ts
```diff
@@ -0,0 +1,107 @@
...
+    date: string;
+    date_formatted: string;
+    hours: string;
+    iso_timestamp: string;
+    minutes: string;
+    proper_iso_timestamp: string;
+    proper_unix_timestamp: number;
+    time: number;
+    timestamp: string;
+    unix_timestamp: number;
```
johnclary: might be good to make these special knack objects' properties optional as well if we're not sure if they are always defined 🤔

johnclary (commented):
Looks great! stuff is rendering! I think it might worth it to tighten up the Knack types/interfaces to err on the side of caution.

🚢

signs-work-order-map/utils/mapUtils.tsx
johnclary: Oops—I left this comment before you updated this. Looks great now!

signs-work-order-map/app/page.tsx
chiaberry: I will rename some things because I agree that this is confusing. I had been visualizing these as signs because that is what they are called in the existing code, but when Christina mentioned multiple assets per location, I got thrown off by what these point meant. 

Location here is either the users's geolocation or could be the location that someone enters manually, but I need to rethink how that is working as I start on https://github.com/cityofaustin/atd-data-tech/issues/21823

signs-work-order-map/types/map.ts
chiaberry: Since I haven't implemented the "add location" functionality yet, this helped me uncover that it is possible in knack to submit and save an empty/blank location (with undefined coordinates) and Knack will save it without any error.

signs-work-order-map/components/Map.tsx
chiaberry: Thank you Mike!

signs-work-order-map/components/SignPopup.tsx
chiaberry: I spun this comment off into its own issue https://github.com/cityofaustin/atd-data-tech/issues/23200, could be a good place to keep track of other tablet usability suggestions or requirements

mddilley (approved):

chiaberry:
I am going to merge this to the collector branch, I will address https://github.com/cityofaustin/atd-knack-signs-markings/pull/327#discussion_r2155288607 in future work and I don't want to have too many differing branches floating out there

PR #328, opened 2025-06-23
Ability for map to load to current location using device location

chiaberry:
https://github.com/cityofaustin/atd-data-tech/issues/21822

This PR adds geolocation to the map. If there are no sign locations on the work order, the map should center at the user's location. If there are sign locations, then the map should fit the signs in view and not zoom to the user's location.

Previously, the javascript code in the iframeMessenger file requested the user's location from the browser and then used the iFrame messaging setup to convey that location to the react app. Now we are instead going to request the location from within the react app and skip the iFrame part. 

I have also added Mapbox's `GeolocateControl` so a user can center the map on their location even after the map first loads. 

### Testing instructions

sign in as a non-coa user, `test-tech@austintexas.gov`, the password is in 1pw under SMD test technician

Pick a work order without locations, for [example](https://atd.knack.com/test-30-may-2024-signs-and-markings-operations#work-order-signs/view-work-orders-details-sign/664e15bd565e0100284b7df5/)
As long as you have location permissions turned on, the map should center on your location. 

Pick a work order with locations, [this one](https://atd.knack.com/test-30-may-2024-signs-and-markings-operations#work-order-signs/view-work-orders-details-sign/5d3f0fa2fa75350013d8b90e/) or [the one from the last PR](https://atd.knack.com/test-30-may-2024-signs-and-markings-operations#work-order-signs/view-work-orders-details-sign/5fa5b8264416d10019334024/). The map does not center at your location, but centers around the signs. 

Test out the GeolocateControl button. Do you think it will be useful to technicians?
Screenshot 2025-06-23 at 5 26 49 PM

Adding locations still isn't in the scope of this PR, and knack doesn't prevent a user from saving an empty location form. But I added some checks so that the app does not crash if the locations are malformed.

### DIFF
--- a/knack/iframeMapMessenger.js
+++ b/knack/iframeMapMessenger.js
@@ -39,13 +39,6 @@
     };
   }
 
-  function AutozoomSendMessageToApp(message) {
-    var iframe = document.getElementById("mapIFrame").contentWindow;
-    const stringifiedMessage = JSON.stringify(message);
-    console.log("autozoom message to app ", message.message);
-    iframe.postMessage(stringifiedMessage, "*");
-  }
-
   // Start polling...
   checkReady(function ($) {
     var $viewSelector = $(myView);
@@ -65,7 +58,7 @@
      */
     function sendMessageToApp(message, iframe) {
       var stringifiedMessage = JSON.stringify(message);
-      iframe.postMessage(stringifiedMessage, "*");
+      iframe.postMessage(stringifiedMessage, nextAppUrl);
     }
 
     // Listen for lat/lon changes
@@ -231,23 +224,5 @@
       sendLocationMapMessage("view_2682");
     });
 
-    // Get the current location from browser.
-    navigator.geolocation.getCurrentPosition(function (position) {
-      // create message object for React App
-      const geolocationMessage = {
-        message: "KNACK_GEOLOCATION",
-        payload: {
-          geolocation: {
-            latitude: position.coords.latitude,
-            longitude: position.coords.longitude,
-          },
-        },
-      };
-
-      // sends geolocation once the iframe is loaded
-      $("#mapIFrame").on("load", function () {
-        AutozoomSendMessageToApp(geolocationMessage);
-      });
-    });
   });
 })();
--- a/knack/iframeMapMessenger.js
+++ b/knack/iframeMapMessenger.js
@@ -7,9 +7,9 @@
 (function () {
   var myView = window.viewIdsArray.shift(0);
 
-  const nextAppUrl =
-    "https://deploy-preview-328--nextjs-knack-signs-markings.netlify.app/";
-  // const nextAppUrl = "http://localhost:3000";
+  // const nextAppUrl =
+  //   "https://deploy-preview-328--nextjs-knack-signs-markings.netlify.app/";
+  const nextAppUrl = "http://localhost:3000";
 
   // Import jQuery into this file from CDN
   // https://stackoverflow.com/questions/34338411/how-to-import-jquery-using-es6-syntax
@@ -62,22 +62,18 @@
     }
 
     // Listen for lat/lon changes
-    // expects a message named "LAT_LON_FIELDS"
+    // expects a message named "LAT_LON_UPDATE"
     // uses lat and lng from message to populate input fields in knack
     window.addEventListener("message", function (event) {
       if (event.origin !== nextAppUrl) {
         return;
       }
       var data = event.data;
-      if (data.message === "LAT_LON_FIELDS") {
-        console.log("received message ", data);
+      if (data.message === "LAT_LON_UPDATE") {
+        console.log("knack received message ", data);
         var $latLonFields = $("#kn-input-field_3300");
-        if (!!data.lat && !!data.lng) {
-          $latLonFields.find("#latitude").val(data.lat);
-          $latLonFields.find("[name='longitude']").val(data.lng);
-        } else {
-          console.error("Payload missing complete location data ", data);
-        }
+        $latLonFields.find("#latitude").val(data.lat);
+        $latLonFields.find("[name='longitude']").val(data.lng);
       }
     });
 
--- a/knack/iframeMapMessenger.js
+++ b/knack/iframeMapMessenger.js
@@ -7,9 +7,9 @@
 (function () {
   var myView = window.viewIdsArray.shift(0);
 
-  const nextAppUrl =
-    "https://deploy-preview-327--nextjs-knack-signs-markings.netlify.app/";
-  // const nextAppUrl = "http://localhost:3000";
+  // const nextAppUrl =
+  // "https://deploy-preview-327--nextjs-knack-signs-markings.netlify.app/";
+  const nextAppUrl = "http://localhost:3000";
 
   // Import jQuery into this file from CDN
   // https://stackoverflow.com/questions/34338411/how-to-import-jquery-using-es6-syntax
@@ -53,8 +53,8 @@
     // Add React app as iframe if iframe doesn't already exist
     if ($(myView + " #mapIFrame").length === 0) {
       https: $(
-        `<iframe src=${nextAppUrl} frameborder="0" scrolling="yes" id="mapIFrame" \
-        style="width: 100%;height: 523px;"></iframe>`
+        `<iframe src=${nextAppUrl} frameborder="0" allow="geolocation" scrolling="yes" \
+        id="mapIFrame" style="width: 100%;height: 523px;"></iframe>`
       ).appendTo($viewSelector);
     }
 
@@ -79,9 +79,12 @@
       if (data.message === "LAT_LON_FIELDS") {
         console.log("received message ", data);
         var $latLonFields = $("#kn-input-field_3300");
-
-        $latLonFields.find("#latitude").val(data.lat);
-        $latLonFields.find("[name='longitude']").val(data.lng);
+        if (!!data.lat && !!data.lng) {
+          $latLonFields.find("#latitude").val(data.lat);
+          $latLonFields.find("[name='longitude']").val(data.lng);
+        } else {
+          console.error("Payload missing complete location data ", data);
+        }
       }
     });
 
--- a/knack/iframeMapMessenger.js
+++ b/knack/iframeMapMessenger.js
@@ -7,9 +7,9 @@
 (function () {
   var myView = window.viewIdsArray.shift(0);
 
-  // const nextAppUrl =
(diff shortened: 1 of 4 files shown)


signs-work-order-map/components/Map.tsx
```diff
@@ -43,9 +48,27 @@ export default function Map({ location, signs }: MapProps) {
...
+
+  /**
+   * If there are no location pins and we have a geolocation point
+   * center map at geolocation
+   */
+  useEffect(() => {
+    if (!mapRef?.current || signs.length > 0) {
+      return;
+    }
+    if (geoLocation?.latitude && geoLocation.longitude) {
```
johnclary: should it be `geoLocation?.longitude` as well? or i guess no chaining needed according to the type def?

knack/iframeMapMessenger.js
```diff
@@ -79,9 +72,12 @@
       if (data.message === "LAT_LON_FIELDS") {
```
johnclary: could this have a more descriptive name? e.g. `LAT_LON_UPDATE` or `SIGN_LOCATION_UPDATE`

knack/iframeMapMessenger.js
```diff
@@ -227,24 +223,5 @@
     $("#view_2682 #mapIFrame").on("load", function () {
       sendLocationMapMessage("view_2682");
     });
-
```
johnclary, line 230: I like that this has moved into the nextjs app 🚀

knack/iframeMapMessenger.js
```diff
@@ -79,9 +72,12 @@
       if (data.message === "LAT_LON_FIELDS") {
         console.log("received message ", data);
         var $latLonFields = $("#kn-input-field_3300");
-
-        $latLonFields.find("#latitude").val(data.lat);
-        $latLonFields.find("[name='longitude']").val(data.lng);
+        if (!!data.lat && !!data.lng) {
+          $latLonFields.find("#latitude").val(data.lat);
+          $latLonFields.find("[name='longitude']").val(data.lng);
```
johnclary: i am wondering why you need a different accessor for lat and lon, but am also too lazy to find out for myself 🙈

knack/iframeMapMessenger.js
```diff
@@ -79,9 +72,12 @@
...
+        } else {
+          console.error("Payload missing complete location data ", data);
```
johnclary: when/why would this happen?

signs-work-order-map/components/Map.tsx
```diff
@@ -82,19 +105,24 @@ export default function Map({ location, signs }: MapProps) {
...
       {popupInfo && (
         <SignPopup popupInfo={popupInfo} setPopupInfo={setPopupInfo} />
       )}
 
       <GeocoderControl position="top-left" marker={true} />
+      <GeolocateControl
+        position="top-left"
+        showUserLocation={false}
+        fitBoundsOptions={{ maxZoom: 16, duration: 0 }}
+      />
```
johnclary, line 124: When I click this button I get a console error. Not sure if it has anything to do with my location being out of the map bounds 🤔 

Screenshot 2025-06-26 at 2 32 06 PM

signs-work-order-map/utils/mapUtils.tsx
```diff
@@ -12,10 +12,17 @@ import { Sign, KnackToIFrameMessage, LatLon } from "@/types/map";
  */
 export const useFormatBounds = (signs: Sign[]): undefined | LngLatBoundsLike =>
   useMemo(() => {
-    if (signs.length === 0) {
+    // Knack will save undefined latitudes and longitudes, this filters those out.
+    const checkedSigns = signs.filter((sign: Sign) => sign.lat && sign.lng);
```
johnclary: The interface definition for `Sign` still indicates that `lat` and `lng` will never be undefined. Would you mind updating that?

johnclary (commented):
This looks good! And I'm glad to see more code moved out of Knack and into the nextjs app 🚀 

I am holding off on approval because I did run into a possible bug.

signs-work-order-map/components/Map.tsx
```diff
@@ -82,19 +105,24 @@ export default function Map({ location, signs }: MapProps) {
...
-        console.log("location from payload: ", location)
-        // when do we show location vs signs? when there are no signs?
-        /* {location?.latitude && location?.longitude && (
-        <Marker latitude={location.latitude} longitude={location.longitude} />
-      )} */
+        // the add location marker, work will be completed in a following PR
+        // location?.latitude && location?.longitude && (
+        //   <Marker latitude={location.latitude} longitude={location.longitude} />
+        // )
+        console.log(location)
```
johnclary: if we're going to keep this log around, would you add some more detail/context to the message?

signs-work-order-map/components/Map.tsx
chiaberry: I was wondering about this! what it was going to do with you being east coast elite

chiaberry: Does it geolocate correctly without the button?

knack/iframeMapMessenger.js
chiaberry: I am going to remove this because I dont think it will ever happen. When you were clicking the "Add location" in previous PRs, I thought it was a failure in the iframe message communication, but the failure was the NextJS app wasnt even attemping to send knack a message and knack was saving empty fields as a location.

signs-work-order-map/components/Map.tsx
johnclary: Without the button and using my real location, the map pans to the edge of the max bounds for the map. I was able to use Chrome dev tools to override my location with a location in Austin, and it worked as expected 👍 

When pushing the button I did run into this error with my real location and the location override 🤷

knack/iframeMapMessenger.js
chiaberry: Never thought to investigate why it was written like this, but now my interest is piqued so 
Screenshot 2025-06-27 at 10 15 42 AM

signs-work-order-map/utils/mapUtils.tsx
chiaberry: I tried doing that and it surfaced errors with the lineString function invocation because it doesnt accept undefined. So I am rewriting instead to filter out the undefined from the payload, so the Sign interface stays as it is, which feels more correct anyway.

knack/iframeMapMessenger.js
```diff
@@ -39,22 +39,15 @@
...
   // Start polling...
   checkReady(function ($) {
     var $viewSelector = $(myView);
 
     // Add React app as iframe if iframe doesn't already exist
     if ($(myView + " #mapIFrame").length === 0) {
       https: $(
-        `<iframe src=${nextAppUrl} frameborder="0" scrolling="yes" id="mapIFrame" \
-        style="width: 100%;height: 523px;"></iframe>`
+        `<iframe src=${nextAppUrl} frameborder="0" allow="geolocation" scrolling="yes" \
```
mddilley, line 49: TIL! This made me curious about why we didn't allow geolocation in the iframe before, and I noticed that this could have [compatibility issues with Safari](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/iframe#browser_compatibility) so maybe it wasn't as compatible with other browsers back when this was first written. 🤷 I don't think this app is used through Safari so looks good!

signs-work-order-map/components/Map.tsx
mddilley: Adding on to say that I didn't see this error message when I tested.

mddilley (approved):
Tested this locally, and I see it working just as you described. I agree that it is nice to see the geolocation code moved closer to the native Mapbox feature/button. I like that the geolocate button is available to users! 📍🚢

signs-work-order-map/components/Map.tsx
johnclary: 👍 probably not worth worrying about in that case

knack/iframeMapMessenger.js
johnclary: 😵

signs-work-order-map/components/Map.tsx
```diff
@@ -43,9 +48,27 @@ export default function Map({ location, signs }: MapProps) {
   });
 
   const bounds = useFormatBounds(signs);
-
   const signPins = useCreateSignPins(signs, setPopupInfo);
+  const geoLocation = useGeoLocation();
+
+  /**
+   * If there are no location pins and we have a geolocation point
+   * center map at geolocation
```
roseeichelmann, line 56: maybe this comment could specify "If there are no sign locations", i was unsure what "location pin" meant at first

signs-work-order-map/components/Map.tsx
```diff
@@ -82,19 +105,24 @@ export default function Map({ location, signs }: MapProps) {
...
 
       {popupInfo && (
         <SignPopup popupInfo={popupInfo} setPopupInfo={setPopupInfo} />
       )}
 
       <GeocoderControl position="top-left" marker={true} />
+      <GeolocateControl
+        position="top-left"
+        showUserLocation={false}
+        fitBoundsOptions={{ maxZoom: 16, duration: 0 }}
```
roseeichelmann: I feel like either the geolocate button or the automatic loading to your location in the `useEffect` could have their zoom adjusted to be more similar? The geolocate button takes me to a really zoomed out version of my location, whereas my location on screen load when there are so signs is really zoomed in. Could take or leave this tho!

signs-work-order-map/utils/mapUtils.tsx
```diff
@@ -52,14 +49,26 @@ export const useFormatSignsRecords = (
...
-    return knackPayload.payload.records.map((sign) => ({
-      id: sign.id,
-      lat: sign.field_3300_raw.latitude,
-      lng: sign.field_3300_raw.longitude,
-      spatialId: sign.field_3297,
-      workOrderId: knackPayload.payload.workOrderId,
-      isLocationDetailPage: sign.id === locationId,
-    }));
+    // Knack will save undefined latitudes and longitudes, this filters those out.
+    const signsArray: Sign[] = knackPayload.payload.records.reduce(
```
roseeichelmann, line 53: :sparkles:

roseeichelmann (approved):
the code looks great and works great, i think the geolocate button will be very helpful. my comments are just suggestions but not stopping this from shipping!! :ship:

signs-work-order-map/components/Map.tsx
chiaberry: good call! I pushed an update to make them match

PR #329, opened 2025-06-26
Ability to create new locations

chiaberry:
https://github.com/cityofaustin/atd-data-tech/issues/21823

The Add Location button appears when a Work Order has the status "Issued".  In production there is a pulsing red marker that appears, in this PR I am using a red marker at an angle. [ Here](https://github.com/cityofaustin/atd-knack-signs-markings/blob/staging/src/index.css#L277) are the styles that in production, I want to get confirmation from Christina that  we should still use the pulsing red marker. The location detail pages use a red marker to signify which location is being viewed and I don't know if that is too similar to the add location marker. 

### Testing instructions

Go to a work order with the status ISSUED. 
Try adding a location
1. Based on your geolocation that happens on map load/ the default map location
2. Dragging the map around to a place and then adding location
3. using the geolocate input (with only basic addresses until https://github.com/cityofaustin/atd-data-tech/issues/21953 is addressed)
4.  Clicking the geolocate button

The page will reload and your new location will be in the locations table immediately above the iframe map.

### DIFF
--- a/signs-work-order-map/app/page.tsx
+++ b/signs-work-order-map/app/page.tsx
@@ -9,14 +9,14 @@ export default function Home() {
 
   console.log(knackPayload);
 
-  const location = useFormatLocation(knackPayload);
+  // const location = useFormatLocation(knackPayload);
   const signs = useFormatSignsRecords(knackPayload);
 
   return (
     <div className={styles.page}>
       <main className={styles.main}>
         <div className={styles.map}>
-          <Map location={location} signs={signs} />
+          <Map signs={signs} messageType={knackPayload?.message}/>
         </div>
       </main>
     </div>
--- a/signs-work-order-map/app/page.tsx
+++ b/signs-work-order-map/app/page.tsx
@@ -2,12 +2,10 @@
 import styles from "./page.module.css";
 import Map from "@/components/Map";
 import { useIFrameMessenger } from "@/utils/iFrameMessenger";
-import { useFormatSignsRecords, useFormatLocation } from "@/utils/mapUtils";
+import { useFormatSignsRecords } from "@/utils/mapUtils";
 
 export default function Home() {
   const knackPayload = useIFrameMessenger();
-
-  console.log(knackPayload);
 
   // const location = useFormatLocation(knackPayload);
   const signs = useFormatSignsRecords(knackPayload);
--- a/signs-work-order-map/app/page.tsx
+++ b/signs-work-order-map/app/page.tsx
@@ -16,7 +16,7 @@ export default function Home() {
     <div className={styles.page}>
       <main className={styles.main}>
         <div className={styles.map}>
-          <Map signs={signs} messageType={knackPayload?.message}/>
+          <Map signs={signs} messageType={knackPayload?.message} />
         </div>
       </main>
     </div>
--- a/signs-work-order-map/components/Map.tsx
+++ b/signs-work-order-map/components/Map.tsx
@@ -4,9 +4,10 @@ import MapGL, {
   MapRef,
   Marker,
   ViewStateChangeEvent,
+  NavigationControl,
+  GeolocateControl,
 } from "react-map-gl/mapbox";
 import GeocoderControl from "@/components/MapGeocoderControl";
-import { NavigationControl, GeolocateControl } from "react-map-gl/mapbox";
 import SignPopup from "./SignPopup";
 import { sendLatLonToParent } from "@/utils/iFrameMessenger";
 import { MapProps, LatLon, Sign } from "@/types/map";
@@ -38,6 +39,18 @@ export default function Map({ signs, messageType }: MapProps) {
     const longitude = +event.viewState.longitude.toFixed(
       MAP_COORDINATE_PRECISION
     );
+    setMapLatLon({
+      latitude,
+      longitude,
+    });
+
+    sendLatLonToParent({ latitude, longitude });
+  }, []);
+
+  const onGeolocate = useCallback((data: GeolocationPosition) => {
+    // truncate values to our preferred precision
+    const latitude = +data.coords.latitude.toFixed(MAP_COORDINATE_PRECISION);
+    const longitude = +data.coords.longitude.toFixed(MAP_COORDINATE_PRECISION);
     setMapLatLon({
       latitude,
       longitude,
@@ -138,6 +151,7 @@ export default function Map({ signs, messageType }: MapProps) {
         position="top-left"
         showUserLocation={false}
         fitBoundsOptions={{ maxZoom: 16, duration: 0 }}
+        onGeolocate={onGeolocate}
       />
       <NavigationControl position="bottom-right" showCompass={false} />
     </MapGL>
--- a/signs-work-order-map/components/Map.tsx
+++ b/signs-work-order-map/components/Map.tsx
@@ -21,6 +21,12 @@ import {
   MAP_COORDINATE_PRECISION,
 } from "@/config/map";
 
+/**
+ *
+ * @param signs Array of Signs from knack paylod, or empty array
+ * @param messageType String from knack payload
+ * @returns
+ */
 export default function Map({ signs, messageType }: MapProps) {
   const mapRef = useRef<MapRef>(null);
   const [popupInfo, setPopupInfo] = useState<Sign | null>(null);
@@ -50,8 +56,8 @@ export default function Map({ signs, messageType }: MapProps) {
   const geoLocation = useGeoLocation();
 
   /**
-   * If there are no location pins and we have a geolocation point
-   * center map at geolocation. Set add location marker to same coords as geolocation
+   * If there are no location pins and we have a geolocation point center
+   * map at geolocation. Set add location marker to same coordindates as geolocation
    */
   useEffect(() => {
     if (!mapRef?.current || signs.length > 0) {
@@ -71,7 +77,7 @@ export default function Map({ signs, messageType }: MapProps) {
 
   /**
    * Zoom to bounding box containing location pins
-   * and set add location marker to center
+   * and set "add location marker" coordinates to center
    */
   useEffect(() => {
     if (!mapRef?.current || !bounds) {
@@ -119,7 +125,6 @@ export default function Map({ signs, messageType }: MapProps) {
         )}
 
       {signPins}
-
       {popupInfo && (
         <SignPopup popupInfo={popupInfo} setPopupInfo={setPopupInfo} />
       )}
--- a/signs-work-order-map/components/Map.tsx
+++ b/signs-work-order-map/components/Map.tsx
@@ -133,11 +133,7 @@ export default function Map({ signs, messageType }: MapProps) {
         <SignPopup popupInfo={popupInfo} setPopupInfo={setPopupInfo} />
       )}
 
-      <GeocoderControl
-        position="top-left"
-        marker={false}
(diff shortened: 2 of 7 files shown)


signs-work-order-map/app/page.tsx
```diff
@@ -2,21 +2,19 @@
...
-import { useFormatSignsRecords, useFormatLocation } from "@/utils/mapUtils";
+import { useFormatSignsRecords } from "@/utils/mapUtils";
 
 export default function Home() {
   const knackPayload = useIFrameMessenger();
 
-  console.log(knackPayload);
-
-  const location = useFormatLocation(knackPayload);
+  // const location = useFormatLocation(knackPayload);
```
chiaberry, line 10: this will come back into use when I tackle https://github.com/cityofaustin/atd-data-tech/issues/23268 next

signs-work-order-map/app/page.tsx
```diff
@@ -2,21 +2,19 @@
...
   const signs = useFormatSignsRecords(knackPayload);
 
   return (
     <div className={styles.page}>
       <main className={styles.main}>
         <div className={styles.map}>
-          <Map location={location} signs={signs} />
+          <Map signs={signs} messageType={knackPayload?.message} />
```
johnclary, line 17: I'm still struggling with the vocab—signs means means "Location" in the knack app and "location" means geolocation.

johnclary (commented):
I can add a location!

I don't mind the red marker (pulsing or no) to indicate the actively edited marker, but I think the UX needs tweaking for this to be intuitive.

When I first clicked **Add location**, I was expecting that to enable me to drag the cursor and then save the location. So on my first try I created a location in the center of the map, where a location already existed. 

I'd prefer that clicking the "Add location" causes the red marker to appear, and that the red marker is not tilted at an angle and alwayss appear on top of all other locations on the map.

The label on the **Add location** button would change to **Save location**, and clicking that button would causes the save action. There should also be a cancel button to exit the edit mode.

I also wonder if, upon saving the location, the page should redirect to the location details page. If I create a location, do I want to immediately add a sign there? In any case, it might be nice if the map zoom/pan persisted after saving the location rather than resetting. 

A few other observations:

* When you click the **Add location** button, you can't see the loading spinner because the button becomes transparent. 

Screenshot 2025-07-14 at 12 08 03 PM

* I know that edit location is a future enhancement, but wanted to mention that it's pretty confusing to see two markers on the map on the location details page. There's no other context on the page that would indicate why that is.

Screenshot 2025-07-14 at 12 13 59 PM

* I don't know if tech's use the geolocate button (which is disabled in my screenshot), but it'd probably be nice to make that larger 

Screenshot 2025-07-14 at 12 21 26 PM

* Showing the list of locations in the Knack table with the map under it is kind of meh, since you can also use the map to navigate to each location details. I wonder about adding a sidebar to the map that you can use to navigate to different locations and doing away with that table entirely. It wold be cool if, when techs are doing this in the field (which they are, right?), the app could be fullscreen with all of the info they need to add locations and sign assets. 

Screenshot 2025-07-14 at 12 24 16 PM

signs-work-order-map/app/page.tsx
chiaberry: https://github.com/cityofaustin/atd-knack-signs-markings/pull/330

chiaberry:
Thanks for testing @johnclary, you gave a lot of great UX feedback. This PR is attempting to replicate what exists in prod now (except the pulsing marker), but the feedback you are giving could be good future knack enhancements, cc: @ChristinaTremel 


> * I know that edit location is a future enhancement, but wanted to mention that it's pretty confusing to see two markers on the map on the location details page. There's no other context on the page that would indicate why that is.
> 

If there are multiple locations on a work order, the one that you are viewing the details of has a red marker and the other locations are blue. I think its confusing UX, which is why I tried putting the new location marker at an angle. Since this one is vertical, it makes me think this work order has two locations wildly far apart from each other. I tried searching for this work order based on the address, but I cant find it.


> * I don't know if tech's use the geolocate button (which is disabled in my screenshot), but it'd probably be nice to make that larger

the geolocate button is new to the work, I don't know if they will want to use the button either! I will see about increasing its size.

signs-work-order-map/app/page.tsx
chiaberry: I said I was going to refactor signs to be called locations, but since I've been in this so long (and comparing to existing prod code) the word `signs` made more sense to me. I'm happy to make a branch without any functionality changes, and just going with location location location to see if it makes more sense to everyone

ChristinaTremel (commented):
Oooo I've been testing creating new locations and really like the functionality! ✨ The accuracy of the location that's created from the pin is great and I like the update made to the pin (in production it's a point and a pin). 
   - Current Map
      image

   - Updated Map
      image

The only buggy thing I am noticing is that the pin used to create the new location isn't staying centered as consistently as what's currently in production. Here's a video of navigating around the map and it seems that I can lose the pin pretty easily. The fix is also easy because if I click on the map again it recenters itself, but I feel like this could potentially cause confusion in the field. 

https://github.com/user-attachments/assets/f7236bb9-1061-4983-bf67-47253ba12142

frankhereford (approved):
Howdy Chia -- this is looking good! 

I have not closely followed others suggestions and comments, but I have enjoyed using the demo you provided in the PR writing up. I'll review other's input next. This enhancement works well, and I think this completes the issue fully. 🚢 ✅ 

I do have some feedback though, and I'm not sure where the line is between making this work like the previous app and enhancing the process, so please feel to spin-off or disregard these ideas -- whatever is best.

* The when you use the marker to save a location, it's saving the location under the white dot on the marker and NOT under the sharp part of the marker's shape. This is going to make 99.99% of the input that is made with the map slightly off, and the amount that the marker will be wrong is based on the zoom level of the map at input time. I think this is very important to fix. 
* Maybe it's a Frank thing, but the "drag-the-map-under-a-stationary-marker" is an unusual interaction for adding a location to me. I see a "click-the-map-to-drop-a-pin" interaction more commonly. 
* Since the marker is relocated `onPan`, can you have it also be relocated `onZoom`? I am hopeful this will eliminate how the cursor can be lost off the edge of the map when zooming in. This issue goes away in a `click to drop a pin` interaction fwiw.
* The marker being tipped over to the right - I was skeptical about this at first, but it has grown on me for when the marker is located behind another marker. However, I think making the red marker always in front of the other markers is clearer, and this would let you tip it back upright which might look more conventional. 
* I think there needs to be some more clear telegraphing of how the "Add Location" functionality works. Maybe one of the two following options would be more clear. I prefer the first one, i think:

  * Start out with no red location in the center. 
  * The user clicks "Add location" to start the process. 
  * The red location appears in the center, and the button becomes "Save location." 
  * When the user clicks "Save location" the mutation is fired off and saved to knack. 
  * When saved, the red location becomes grey and the button goes back to "Add Location"

OR

* Rename the button to "Save Current Location" and if you wanted to really express to the user that we're talking about the location under the red marker, you might add an icon of the red marker to the left of "Save Cur.." in the label. 
  * A last idea, for better or worse: Or add some tiny latitude and longitude numbers under that label on the button that update `onMove` -- not intended to give people access to the specific digits, but to rather show them that they change when they interact with the map, teaching them how to use the interaction.

frankhereford (changes requested):
Howdy Chia -- this is looking good! 

I have not closely followed others suggestions and comments, but I have enjoyed using the demo you provided in the PR writing up. I'll review other's input next. This enhancement works well, and I think this completes the issue really well.

There is one thing that I feel pretty strongly about, but if it's not doable, please let me know and I'll turn this to a ✅ . What do you think about this?

* The when you use the marker to save a location, it's saving the location under the white dot on the marker and NOT under the sharp part of the marker's shape. This is going to make 99.99% of the input that is made with the map slightly off, and the amount that the marker will be wrong is based on the zoom level of the map at input time. I think this is worth fixing because it will have a meaningful impact on data accuracy.

----

Other than that, I have some general feedback, and I'm not sure where the line is between making this work like the previous app and enhancing the process, so please feel to spin-off or disregard these ideas -- whatever is best.


* Maybe it's a Frank thing, but the "drag-the-map-under-a-stationary-marker" is an unusual interaction for adding a location to me. I see a "click-the-map-to-drop-a-pin" interaction more commonly. 
* Since the marker is relocated `onPan`, can you have it also be relocated `onZoom`? I am hopeful this will eliminate how the cursor can be lost off the edge of the map when zooming in. This issue goes away in a `click to drop a pin` interaction fwiw.
* The marker being tipped over to the right - I was skeptical about this at first, but it has grown on me for when the marker is located behind another marker. However, I think making the red marker always in front of the other markers is clearer, and this would let you tip it back upright which might look more conventional. 
* I think there needs to be some more clear telegraphing of how the "Add Location" functionality works. Maybe one of the two following options would be more clear. I prefer the first one, i think:

  * Start out with no red location in the center. 
  * The user clicks "Add location" to start the process. 
  * The red location appears in the center, and the button becomes "Save location." 
  * When the user clicks "Save location" the mutation is fired off and saved to knack. 
  * When saved, the red location becomes grey and the button goes back to "Add Location"

OR

* Rename the button to "Save Current Location" and if you wanted to really express to the user that we're talking about the location under the red marker, you might add an icon of the red marker to the left of "Save Cur.." in the label. 
  * A last idea, for better or worse: Or add some tiny latitude and longitude numbers under that label on the button that update `onMove` -- not intended to give people access to the specific digits, but to rather show them that they change when they interact with the map, teaching them how to use the interaction.

signs-work-order-map/components/Map.tsx
```diff
@@ -91,36 +121,39 @@ export default function Map({ location, signs }: MapProps) {
       }}
       cooperativeGestures={true}
       {...DEFAULT_MAP_PARAMS}
-      onDrag={onDrag}
+      onDrag={updateCenterMarker}
+      onZoom={updateCenterMarker}
+      onMoveEnd={updateCenterMarker}
```
chiaberry, line 126: updated based on feedback, listening now for also changes on zoom and `onMoveEnd` is to handle the behavior Christina found where if you drag and then let go and the map kind of drifts, the marker doesn't follow.

signs-work-order-map/utils/mapUtils.tsx
```diff
@@ -111,7 +112,6 @@ export const useCreateSignPins = (
               key={`marker-${sign.id}`}
               longitude={sign.lng}
               latitude={sign.lat}
-              anchor="bottom"
```
chiaberry, line 114: removing this seems to have fixed the markers not behaving correctly when the map was zooming in and out

chiaberry:
@ChristinaTremel and @frankhereford this is ready for eyes again, to see if the marker behavior has improved

ChristinaTremel (approved):
I've had an opportunity to test out creating new locations after the changes you've made @chiaberry and the functionality of the map feels way smoother! ✨  

I also got to dive deeper into the comments from Frank and John and really appreciate the feedback and ideas to improve the map! I think the technicians are super familiar with the usability of the map from the perspective of the point staying centered and moving the map around so I don't want to change that up on them. But I'm super interested in the comment about maybe changing the way the locations table is formatted.. I think we can hold off on any changes to the location table in this PR but we might want to look at that functionality when refine #23268.

chiaberry:
@frankhereford I am going to go ahead and merge this to the 21517-app-next branch so @mateoclarke can have this code in his feature branch, please let me know though if my updates did not address the marker location issue you encountered

PR #330, opened 2025-07-14
Ability to edit existing knack locations

chiaberry:
In production, only GIS Technicians see the edit location link on a location details page - 

### Testing instructions

to see these changes, update [the file in s3 ](https://atd-knack-signs-markings.s3.us-east-1.amazonaws.com/staging/iframeMapMessengerDev.js)to point to either this PR or localhost. 

Go to a work order with the status ISSUED with existing locations

Click on a location's details icon to see it's location details page.
Screenshot 2025-07-14 at 3 46 12 PM

 at the top there should be an edit link 

Screenshot 2025-07-14 at 3 45 18 PM

The map loads with that location as the center. It's latitude and longitude are in form below the map. You can drag the map to move the marker and the lat/lon update below. "Submit" saves the update.

### DIFF
--- a/signs-work-order-map/components/Map.tsx
+++ b/signs-work-order-map/components/Map.tsx
@@ -70,7 +70,8 @@ export default function Map({ signs, messageType, editLocation }: MapProps) {
 
   /**
    * If there are no sign location pins and we have a geolocation point center
-   * map at geolocation. Set add location marker to same coordindates as geolocation
+   * map at geolocation *unless we are editing an existing saved location*.
+   * Set add location marker to same coordindates as geolocation
    */
   useEffect(() => {
     if (
@@ -115,7 +116,9 @@ export default function Map({ signs, messageType, editLocation }: MapProps) {
   }, [bounds]);
 
   /***
-   *
+   * editLocation is sent from knack when a GIS Tech is on the Edit Location view
+   * center map on location, and send coordinates back to Knack to populate the
+   * location form
    */
   useEffect(() => {
     if (!mapRef?.current || !editLocation) {
--- a/signs-work-order-map/components/Map.tsx
+++ b/signs-work-order-map/components/Map.tsx
@@ -28,7 +28,7 @@ import {
  * @param messageType String from knack payload
  * @returns
  */
-export default function Map({ signs, messageType }: MapProps) {
+export default function Map({ signs, messageType, editLocation }: MapProps) {
   const mapRef = useRef<MapRef>(null);
   const [popupInfo, setPopupInfo] = useState<Sign | null>(null);
   const onDrag = useCallback((event: ViewStateChangeEvent) => {
--- a/signs-work-order-map/components/Map.tsx
+++ b/signs-work-order-map/components/Map.tsx
@@ -73,7 +73,11 @@ export default function Map({ signs, messageType, editLocation }: MapProps) {
    * map at geolocation. Set add location marker to same coordindates as geolocation
    */
   useEffect(() => {
-    if (!mapRef?.current || signs.length > 0) {
+    if (
+      !mapRef?.current ||
+      signs.length > 0 ||
+      messageType === "EDIT_LOCATION"
+    ) {
       return;
     }
     if (geoLocation?.latitude && geoLocation?.longitude) {
@@ -86,7 +90,7 @@ export default function Map({ signs, messageType, editLocation }: MapProps) {
         longitude: geoLocation.longitude,
       });
     }
-  }, [geoLocation, signs]);
+  }, [geoLocation, signs, messageType]);
 
   /**
    * Zoom to bounding box containing location pins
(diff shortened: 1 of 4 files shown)


chiaberry:
I forgot about this PR, I was holding off on requesting reviews until the previous PR had been reviewed and merged and that one sat out there for a while. 

@ChristinaTremel This is the edit location view I was talking about during refinement, wondering if it should be allowed for  locations chosen from the agol dataset too

roseeichelmann (approved):
all the test steps work as expected! 

this may be out of scope but i find it distracting that the map always loads to a different location before updating to be where the markers actually are. ive noticed its usually 15th st or airport blvd. idk if this can be fixed but thought it was worth mentioning

https://github.com/user-attachments/assets/b38dc0fa-78c0-45dd-9091-36160a4bac32

signs-work-order-map/components/Map.tsx
```diff
@@ -111,6 +115,28 @@ export default function Map({ signs, messageType }: MapProps) {
     });
   }, [bounds]);
 
+  /**
+   * editLocation is sent from knack when a GIS Tech is on the Edit Location view
+   * center map on location, and send coordinates back to Knack to populate the
+   * location form
```
frankhereford: It's probably me, but I am stumbling on this comment. I think it can be misread to indicate that a knack service is emitting a HTTP request to ping a webhook or kick off some event off of their network. Maybe:

```
/**
 * editLocation Handler
 * 
 * Triggered when a GIS Technician accesses the Edit Location view in Knack.
 * Centers the map on the current location coordinates and captures user input
 * to update the location form fields with new coordinate data.
 */
```

frankhereford (approved):
Worked great - thanks for this PR! Tested in the knack app and the code reads well. I did get tripped up on a comment, but that couldn't be less of a thing. Thanks!

signs-work-order-map/components/Map.tsx
```diff
@@ -60,21 +64,21 @@ export default function Map({ signs, messageType }: MapProps) {
...
-   * map at geolocation. Set add location marker to same coordindates as geolocation
+   * map at geolocation *unless we are editing an existing saved location*.
+   * Set add location marker to same coordindates as geolocation
    */
   useEffect(() => {
-    if (!mapRef?.current || signs.length > 0) {
+    if (
+      !mapRef?.current ||
+      signs.length > 0 ||
+      messageType === "EDIT_LOCATION"
```
mddilley: It is probably too early to optimize like this, but I wonder if there is a future state where this Map component doesn't need to be aware of the message type at all and would take props that could take an initial lat/lon from the parent components that map to Knack views for example. I'm leaving out a lot of other props that would make this more complex. 

I know this gets tough with different views requiring different capabilities, and I am also far away from this work!

signs-work-order-map/components/Map.tsx
mddilley: I forgot that the `sendLatLonToParent` (`onDrag` handler) was responsible for sending the lat/lon back through the iframe messenger. So maybe adding to this comment about this would be helpful too?

signs-work-order-map/components/Map.tsx
mddilley, line 1: (Commenting on this whole file since this is affected by logic on the Knack side and React side) I also found the jump between the default center to the bounding box with all locations in view to be a bit jarring after saving the updated lat/long, but I also realize that saving the form navigates from the "Edit Location" Knack page back to the "Location Details" Knack page which probably complicates things with the transfer of data over the iframe messaging.

I think what I was missing was visual feedback about the new lat/lon of the location just edited when I arrived back at the "Location Details" page. I know it is a tough balance between being able to see all locations in the map and being able to see the specific location of one.

mddilley (approved):
I tested this using the deploy preview, and I also looked at the code locally, and I was able to save updates to an existing location many times. ✅ I continue to be amazed at seeing this code adapted to Next.js/TypeScript. 🚢 🚀

signs-work-order-map/components/Map.tsx
```diff
@@ -111,6 +115,28 @@ export default function Map({ signs, messageType }: MapProps) {
...
+    }
+
+    if (editLocation.latitude && editLocation.longitude) {
+      mapRef.current.jumpTo({
+        center: [editLocation?.longitude, editLocation?.latitude],
+      });
+
+      setMapLatLon({
+        latitude: editLocation.latitude,
+        longitude: editLocation.longitude,
```
johnclary: do we need the `.toFixed(MAP_COORDINATE_PRECISION)` transform here?

signs-work-order-map/components/Map.tsx
```diff
@@ -111,6 +115,28 @@ export default function Map({ signs, messageType }: MapProps) {
     });
   }, [bounds]);
 
+  /**
+   * editLocation is sent from knack when a GIS Tech is on the Edit Location view
+   * center map on location, and send coordinates back to Knack to populate the
+   * location form
+   */
+  useEffect(() => {
+    if (!mapRef?.current || !editLocation) {
```
johnclary: Is testing `editLocation` different than checking for message type `"EDIT_LOCATION"` above? i am getting a bit confused about the three useeffects in this component and when we want them to run. Perhaps I am echoing others' comments about wanting a bit more clarity in the comments.

johnclary (commented):
I tested this and it works just as promised! Good to be able to edit locations. Thank you for continuing to move this challenging project forward!

Some more UI/UX feedback:
* Like others have commented, I'm not a fan of the jump to the edit location that happens after the map loads. Maybe it's possible to block map rendering until we have all the data we need to load it in the correct location? 
* I know there was a reason for it at some point, but I can't recall why the red location marker is tilted. It looks a bit off. 
* When I load the location details page, the tilted red edit marker sometimes briefly appears (screenshot below)
* I've mentioned this elsewhere, but it would be nice if the **Edit** button were larger, and if there was a more clear visual queue about when you're editing a location versus viewing the details page, since the pages look almost identical
* On my laptop, the **Submit** button is below the fold and I have to scroll to find it. It wold be nice if it floated above the map so i could click it without worrying about messing up my location.
* Minor thing, and maybe just a me problem, but the "Use <command> + scroll to zoom the map" text is oddly small. I suppose this is probably an annoyance related to iframes

Image

Image

PR #332, opened 2025-11-10
Add AGOL sign location assets to map

mateoclarke:
Closes https://github.com/cityofaustin/atd-data-tech/issues/25352

Screenshot 2025-11-09 at 8 57 46 PM

## Testing

You can test this feature in the deploy preview or locally:
https://deploy-preview-332--nextjs-knack-signs-markings.netlify.app/

When you load the app, you should see: 
- yellow dots appear representing signs assets fetched from AGOL
- the console should tell you how many assets were fetch within the bounding box visible on screen
- when you drag the map, you should see new assets fetch
- if you click a yellow dot, a few summary details should print to the dev console.

One area for feedback is the map gets very laggy as you zoom out and fetch too many assets. Perhaps we should hide these at a certain zoom level if you folks have the same experience in testing? I can do that as a follow on PR but wanted to get feedback before the scope grows.


signs-work-order-map/utils/mapUtils.tsx
```diff
@@ -91,10 +95,39 @@ export const useFormatLocation = (
...
+      boxShadow: "0 1px 2px rgba(0,0,0,0.3)", // Subtle shadow for visibility
+      transform: "translate(-50%, -50%)", // Center the dot on the coordinate
+    }}
+  />
+);
+
+/**
+ * Helper function to log AGOL sign data to console
+ */
+const logAGOLSignData = (sign: Sign) => {
```
mateoclarke, line 120: This could be deleted once we implement interactivity with each sign asset point. For now, I just wanted to demonstrate high-level details about each point for future iteration.

signs-work-order-map/components/Map.tsx
```diff
@@ -159,6 +202,45 @@ export default function Map({ signs, messageType, editLocation }: MapProps) {
...
+            background: "rgba(255, 255, 255, 0.9)",
+            padding: "8px 12px",
+            borderRadius: "4px",
+            fontSize: "14px",
+            boxShadow: "0 2px 4px rgba(0,0,0,0.1)",
+            zIndex: 1000,
+          }}
+        >
+          Loading sign assets...
+        </div>
```
frankhereford: I had to throttle down my network to get to actually see this, but I'm glad I did -- it looks good!

signs-work-order-map/utils/agol.ts
```diff
@@ -0,0 +1,254 @@
...
+
+  return {
+    type: "FeatureCollection",
+    features: uniqueFeatures,
+  };
+};
+
+/**
+ * Hook to query AGOL feature service for features within a bounding box
+ * Continuously accumulates features into a single FeatureCollection for seamless map experience
```
frankhereford, line 117: This was my biggest question going into this review; how are you handling getting more features than you can get in a single query, and this handles that really elegantly -- premium solution!

signs-work-order-map/utils/mapUtils.tsx
frankhereford: Or, maybe leave it in and make it use `console.info()` - my motivation for this comment is that the console logging you've got going here is really good in how it conveys events, counts etc..  and is just generally an example of information-dense logging that is also readable at the same time. I'd hate to loose it because i think future us would love to have it stick around when we pick it back up to dev on.

signs-work-order-map/utils/mapUtils.tsx
```diff
@@ -91,10 +95,39 @@ export const useFormatLocation = (
...
+  console.group(`🟡 AGOL Sign Asset - ID: ${sign.spatialId}`);
```
frankhereford, line 121: `group()` -- this is fancy and i really dig it.

signs-work-order-map/utils/mapUtils.tsx
```diff
@@ -91,10 +95,39 @@ export const useFormatLocation = (
...
+  console.log("📍 Location:", `${sign.lat.toFixed(6)}, ${sign.lng.toFixed(6)}`);
```
frankhereford: * 👍 on 6 decimal degrees -- at our latitude, that's about 10 CM, which feels like a good minimum resolution when talking about the location of a sign post that's maybe 4-5 CM in diameter. 

* Maybe instead of the magic `6` in here, is it possible to use the configuration value, with something like `import { MAP_COORDINATE_PRECISION } from "@/config/map";`?

signs-work-order-map/utils/mapUtils.tsx
```diff
@@ -112,14 +145,34 @@ export const useCreateSignPins = (
               key={`marker-${sign.id}`}
               longitude={sign.lng}
               latitude={sign.lat}
-              color={sign.isLocationDetailPage ? "red" : undefined}
+              color={
+                sign.isLocationDetailPage
+                  ? "red"
+                  : sign.source === "agol"
+                    ? undefined // Don't use default color for AGOL signs
+                    : undefined
```
frankhereford: I'm curious - is this second, inner ternary holding a place for a future enhancement? It stuck out since both branches resolve to `undefined`. If not a placeholder, can you reduce this down to just the enclosing ternary?

frankhereford (approved):
This tested out real fine Mateo, and the code is /really/ great. I think your architecture choices with the `useAGOLSignAssets` and `useFeatureService` hooks is total quality.

This is 🚢🚢🚢 -- and as an aside, I was really impressed by the quality of sign data we have. I took my laptop out to peek compare our data to reality, and it's really good based on my tiny sample here in North University.

Thanks Mateo!

signs-work-order-map/utils/agol.ts
mateoclarke: This is all thanks the pattern in Moped that Mike clued me into. The "accumulation" concept is so good and much better than the way I was approaching this at first.

signs-work-order-map/utils/mapUtils.tsx
mateoclarke: This was a TIL for me too. Thanks robot overlords.

signs-work-order-map/utils/mapUtils.tsx
mateoclarke: Yeah good call out. This `undefined` to signify default color is a carryover from the code that existed before, so I didn't scrutinize it much at first. But now its more convoluted. And nested ternaries are a code smell. I'll give this another look before I merge. I think we can clarify.

signs-work-order-map/components/Map.tsx
```diff
@@ -35,6 +37,13 @@ export default function Map({ signs, messageType, editLocation }: MapProps) {
     latitude: DEFAULT_MAP_PAN_ZOOM.latitude,
     longitude: DEFAULT_MAP_PAN_ZOOM.longitude,
   });
+  const [mapBounds, setMapBounds] = useState<{
+    north: number;
+    south: number;
+    east: number;
+    west: number;
+  } | null>(null);
```
mddilley, line 47: I've been enjoying learning about types in map code in here and VZ, and I noticed [LngLatBoundsLike](https://github.com/cityofaustin/vision-zero/blob/35f98c68c8615faf86ce22a97d3e5f8503038cb6/editor/components/MapFitBoundsControl.tsx#L18) was used in VZE as an alternative for map bounds type. All of y'all's work certainly makes the long-term goal of TS-ifying Moped code look less intimidating. 🚀

Edit: Ok now I see the getMapBounds helper where the cardinal directions are populated and how that fits into the type. 💯

signs-work-order-map/components/Map.tsx
mddilley: +1 - this is a nice touch! I've had loading and error states for the Moped AGOL map layer on the brain since our recent outages. Very cool to have feedback about what is happening.

signs-work-order-map/types/map.ts
```diff
@@ -17,6 +17,10 @@ export interface Sign {
   workOrderId: string;
   /** if the sign id matches the location record id, the payload came from the location details page */
   isLocationDetailPage: boolean;
+  /** source of the sign data - 'knack' for existing data, 'agol' for ArcGIS Online */
+  source?: "knack" | "agol";
+  /** additional attributes from AGOL or other sources */
+  attributes?: any;
```
mddilley, line 23: I like using `any` type to keep the door open since we don't know exactly what attributes will be important in this map yet.🚪

Not a change for this PR, but I'm getting confused about our TS config that maybe someone could help me sort out. I thought that using `"strict": true` in our config would enforce the `no-explicit-any` rule and fail our deploy preview builds? Or maybe that is specific to VZ. 🤔

signs-work-order-map/utils/agol.ts
```diff
@@ -0,0 +1,254 @@
...
+  north: number;
+  south: number;
+  east: number;
+  west: number;
+}
+
+interface AGOLFeature {
+  type: "Feature";
+  properties: {
+    [key: string]: any;
```
mddilley, line 36: thinking about `any` vs. `unknown` here since I still get caught up in the difference. But, if we used `unknown`, we would have to write code to check types before accessing a key from the AGOL payload (which could change without us knowing) I think. `any` leaves us without type guards and possible crashing at runtime but is more flexible. I'm still trying to learn so wanted to see if this lines up with how we are using them.

signs-work-order-map/components/Map.tsx
```diff
@@ -65,9 +82,30 @@ export default function Map({ signs, messageType, editLocation }: MapProps) {
...
-  const signPins = useCreateSignPins(signs, setPopupInfo);
   const geoLocation = useGeoLocation();
 
+  // Fetch AGOL sign assets based on current map bounds
+  const {
+    agolSigns,
+    loading: agolLoading,
+    error: agolError,
+  } = useAGOLSignAssets(
+    mapBounds,
```
mddilley, line 102: i like that you called this `mapBounds`! Makes me wonder about updating the existing `bounds` var to `signBounds` in a future PR. I found myself getting turned around while exploring the code locally so wanted to mention.

signs-work-order-map/utils/agol.ts
mddilley: h/t to John for this pattern that has served Moped well for a long time! And, Mateo, I agree with your observation in the PR description about hiding signals when zoomed out to a certain threshold. There is a similar pattern in Moped so we don't fetch a huge number of features with an alert that says "Zoom in to select features."

Probably even more important here since this app is used out in the field on mobile devices since I had noticeable slowdown on my desktop browser when zoomed out. cc @ChristinaTremel

signs-work-order-map/components/Map.tsx
```diff
@@ -159,6 +202,45 @@ export default function Map({ signs, messageType, editLocation }: MapProps) {
...
+            background: "rgba(255, 0, 0, 0.9)",
+            color: "white",
+            padding: "8px 12px",
+            borderRadius: "4px",
+            fontSize: "14px",
+            boxShadow: "0 2px 4px rgba(0,0,0,0.1)",
+            zIndex: 1000,
+          }}
+        >
+          Error loading signs: {agolError}
```
mddilley: Noting that I wasn't able to get this error to display on the map in offline browser mode but also checking in if that is expected.

mddilley (approved):
Awesome to see this map utilizing the signs feature layer! 🚀 I left some notes about possible scope that may be covered in future issues, but this looks ready for the next issue to me. Lots of TILs in here! 🙏 🚢

signs-work-order-map/components/Map.tsx
mateoclarke: I reorganized the message components into their own file and fixed the error state handling with my last commit. Thanks for catching this edge case!
Screenshot 2025-11-17 at 9 02 55 PM

mateoclarke:
Based on [Mike's comment](https://github.com/cityofaustin/atd-knack-signs-markings/pull/332#discussion_r2528284032), I went ahead and added a zoom tracking feature to monitor the zoom threshold, and once it zoom out to a number less than 15, signs are hidden. I could still use some feedback if the signs should be hidden until youre zoomed in more or less. cc/ @ChristinaTremel @chiaberry

ChristinaTremel (changes requested):
Thanks so much for this work, Mateo!! Looks so good. I think I'd like signs to be hidden until we're a little more zoomed into the map. I don't know how to communicate the zoom level but I think if we hid the signs around this point of zoom in the screenshot, that would be great!
image

mateoclarke:
Thanks @ChristinaTremel thats super helpful. I've updated the threshold to 16.75. Here's a few screenshots so you can see that threshold:

## Zoom 16.74 (signs hidden)
Screenshot 2025-11-19 at 9 33 00 PM


## Zoom 16.77 (signs visible)
Screenshot 2025-11-19 at 9 33 07 PM

signs-work-order-map/components/Map.tsx
johnclary, line 1: outside of the scope of this PR, but wanted to mention that it would be nice to update the NextJS version of this app and bump a few other outdated packages.

signs-work-order-map/components/Map.tsx
```diff
@@ -43,14 +55,25 @@ export default function Map({ signs, messageType, editLocation }: MapProps) {
...
       longitude,
     });
+    setZoom(currentZoom);
 
     sendLatLonToParent({ latitude, longitude });
   }, []);
 
+  // Update map bounds when the map moves
+  const updateMapBounds = useCallback(() => {
+    const bounds = getMapBounds(mapRef);
```
johnclary, line 71: i was getting a little confused by all the `bounds` floating around in this file. might be nice to rename the var that comes from `useFormatBounds(signs)` as something like `signLocationBounds` just to make the distinction more clear.

signs-work-order-map/components/Map.tsx
```diff
@@ -65,9 +88,36 @@ export default function Map({ signs, messageType, editLocation }: MapProps) {
...
+    agolSigns,
+    loading: agolLoading,
+    error: agolError,
+  } = useAGOLSignAssets(
+    mapBounds,
+    mapLoaded && isZoomedInEnough // Only fetch when map is loaded and zoomed in enough
+  );
+
+  // Combine Knack signs with AGOL signs (only include AGOL signs if zoomed in enough)
+  const allSigns = useMemo(() => {
```
johnclary, line 107: I think I asked this question in another PR, but IIRC we have up to his point been dealing with sign _location_ records, which have a parent -> child relationship with sign records (multiple signs can be associated with one location).

I think it would be worth the trouble of renaming props and vars in this component to be consistent about the naming conventions and make sure that this app uses the same terminology as the knack app.

johnclary: And just to help my understanding—the technicians are not going to need to interact with individual sign assets in this app, they just need sign locations, right?

i see that the sign location [feature layer](https://services.arcgis.com/0L95CJ0VTaxqcmED/arcgis/rest/services/Sign_Assets_Maint_Public_View/FeatureServer/1) has a `SIGN_MESSAGES_AT_LOCATION` field with comma-separated sign names in it—so maybe the intention is to render that in the app at some point.

```
CELL PHONE USE PROHIBITED UP TO $200 FINE, CELL PHONE USE PROHIBITED UP TO $200 FINE, SCHOOL SPEED LIMIT 25 7:00AM-8:30AM 2:30PM-4:00PM MON-FRI, SCHOOL SPEED LIMIT 25 7:00AM-8:30AM 2:30PM-4:00PM MON-FRI
```

signs-work-order-map/components/MapStatusIndicator.tsx
```diff
@@ -0,0 +1,81 @@
...
+ * {isLoading && (
+ *   <MapStatusIndicator type="loading" message="Loading sign assets..." />
+ * )}
+ *
+ * {error && (
+ *   <MapStatusIndicator type="error" message={`Error: ${error}`} />
+ * )}
+ * ```
+ */
+export const MapStatusIndicator = ({
```
johnclary, line 37: This is a nice touch! Might be good to move the styles to stylesheet. I'm also curious if we want to get fancy with a loading spinner ⌛

signs-work-order-map/config/map.ts
```diff
@@ -5,10 +5,10 @@ import "mapbox-gl/dist/mapbox-gl.css";
 // const NEARMAP_KEY = process.env.NEXT_PUBLIC_NEARMAP_KEY;
 export const MAPBOX_TOKEN = process.env.NEXT_PUBLIC_MAPBOX_TOKEN;
 
-
 // TODO: update to prod url when testing is completed
 // export const KNACK_APP_URL = 'http://atd.knack.com/signs-markings'
-export const KNACK_APP_URL = 'https://atd.knack.com/test-30-may-2024-signs-and-markings-operations';
+export const KNACK_APP_URL =
+  "https://atd.knack.com/test-30-may-2024-signs-and-markings-operations";
```
johnclary, line 11: should we be using an env var for this?

signs-work-order-map/types/map.ts
johnclary: I'm also surprised this isn't failing the linter. I'd like to see us avoid the `any` type if possible. if we're not using the `attributes` at the moment, perhaps `unknown` or `Record<string, unknown>` would suffice?

signs-work-order-map/utils/agol.ts
johnclary: Fun to see this code live on—with some nice improvements and polish 🙌

signs-work-order-map/utils/agol.ts
```diff
@@ -0,0 +1,282 @@
...
+    isLocationDetailPage: false,
+    source: "agol" as const,
+    attributes: feature.properties,
+  }));
+};
+
+/**
+ * Find a specific feature in the GeoJSON collection by ID
+ */
+export const findFeatureById = (
```
johnclary, line 274: wouldn't be surprised if this comes in handy later, but this function does not appear to be in use currently.

signs-work-order-map/utils/mapUtils.tsx
```diff
@@ -91,11 +95,51 @@ export const useFormatLocation = (
...
+      width: "8px",
+      height: "8px",
+      borderRadius: "50%",
+      backgroundColor: "#FFD700", // Gold/Yellow color
+      border: "1px solid #000000", // Black stroke
+      cursor: "pointer",
+      boxShadow: "0 1px 2px rgba(0,0,0,0.3)", // Subtle shadow for visibility
+      transform: "translate(-50%, -50%)", // Center the dot on the coordinate
+    }}
+  />
```
johnclary, line 114: It'd be nice to move these to a stylesheet as well. I assume we'll probably tweak the marker contrast and size for usability later?

It might be worth considering if we want to render the locations on the map via a MapGL `geojson` layer instead of adding `Marker` components. I _think_ there could be a pretty significant render performance improvement because of how many locations there, which might be worth it for whatever devices they're using in the field. The other benefit is that we could make use of dynamic styles based on map zoom and other feature data.

I'm happy to expand on this thought, and I mentioned a similar consideration [here](https://github.com/cityofaustin/vision-zero/pull/1891#discussion_r2511978651) when I made fancy icon markers for the EMS map.

johnclary (approved):
@mateoclarke nice work! very cool to see the map come to life with external data from arcgis 🚀

signs-work-order-map/components/Map.tsx
```diff
@@ -159,6 +218,24 @@ export default function Map({ signs, messageType, editLocation }: MapProps) {
         <SignPopup popupInfo={popupInfo} setPopupInfo={setPopupInfo} />
       )}
 
+      {/* Status indicators for AGOL data */}
+      {!isZoomedInEnough && (
+        <MapStatusIndicator
+          type="loading"
+          message="Zoom in to view sign features"
```
roseeichelmann, line 225: ❤️

roseeichelmann (approved):
this looks great,i like the zoom threshold you added with the little message in the top right corner to zoom in. woohoo 🚀

mateoclarke:
@ChristinaTremel @chiaberry, what's the status on this? Are we ready to do a final review and merge?

signs-work-order-map/types/map.ts
johnclary: Not to sound butthurt about my own feedback not sticking, but I was surprised to see this merge without any more discussion on this or a few other comments I left here 😕

mateoclarke: My apologies @johnclary. I overlooked your feedback from Thanksgiving week, my bad. I just saw your ✅ and hastily assumed I was good to ship. But I'll start a new branch to address the type issues, styling comments, etc. I believe the Next upgrade comment has been addressed by the combo of my last few commits on this PR and [Frank's separate PR](https://github.com/cityofaustin/atd-knack-signs-markings/pull/333). So I'll circle back on the local linter not working in my next PR by branching from @frankhereford. 

On the product-focused decisions, I'll tag Christina and Chia for their visibility.

johnclary: sounds good—thank you!

signs-work-order-map/components/Map.tsx
mateoclarke: @ChristinaTremel @chiaberry, I'm going to tag y'all here to decide if you want to create and assign a new issue to align terminology between this app and knack in regards to `sign locations` and `sign records`. I'm not sure the terminology in the knack app so it feels like this needs more scoping to implement.

Also to field the question about the intention for how technicians interact with sign assets vs sign locations.

signs-work-order-map/config/map.ts
mateoclarke: Is it a secret? This diff was just an autoformatting change on my end, so I'm not inclined to change it. Tagging @chiaberry for visibility.

signs-work-order-map/utils/mapUtils.tsx
mateoclarke: Thanks for sharing the feedback about changing the way we render the locations. This makes sense as a more modern sleek way to handle a bunch of components and I wish I knew about it before tackling this issue. Makes me feel like this issue didn't get the time it deserved in dev backlog refinement/planning to float this solution and include it in the scope.

I went ahead and made an issue for the backlog that maybe @ChristinaTremel and @chiaberry can discuss in a dev/product sync and decide when/whether to bring forward into a sprint. I can definitely see how it would help the performance, especially on mobile devices. What we implemented in this PR works as an MVP and there wasn't too much feedback about performance once I included the bounding box loading, so I don't want to take on the effort of this wholesale change in marker implemenatation without more product input and potentially user feedback through testing.

Here it is for y'alls consideration:
https://github.com/cityofaustin/atd-data-tech/issues/26123

chiaberry: Thanks for making the issue for the backlog, I am going to remove myself from the issue and defer the product decisions and timing to @ChristinaTremel. 

We did discuss this issue in dev backlog refining at length, iirc we spent most of a meeting only on this issue, but it was a lot of scoping of behavior and not as heavy on defining exact implementation steps.

PR #338, opened 2026-01-20
Adds pop-ups for rendered AGOL sign locations

mateoclarke:
Closes https://github.com/cityofaustin/atd-data-tech/issues/26317

Screenshot 2026-01-20 at 10 36 18 AM

## Testing:
1. Go to this deploy preview: https://deploy-preview-338--nextjs-knack-signs-markings.netlify.app
2. Click on a yellow marker representing a sign asset, and verify the pop-up contains a Location ID and a list of 1 or more sign messages.
3. Click the map to close, open again, and click the `X` to close. Open again and click another asset marker to see the pop-up's data update, and that it moves the position of the pop-up, but doesn't close. Basically, just make sure the pop-up open and close behavior works as expected.

PR #339, opened 2026-02-05
Migrates AGOL sign markers to MapGL GeoJSON layer for better performance

mateoclarke:
Closes https://github.com/cityofaustin/atd-data-tech/issues/26123

This PR should have no visual difference except for performance enhancements.

I started by asking Cursor to create a way we could benchmark the performance of the map using individual Marker components vs a GeoJSON layer implementation with increasing amounts of map pins in memory. It created a page at `/benchmark` that helps illustrate how much of a performance difference this makes.  Since each marker is a distinct DOM Element, it really slows down the reactivity of the map, especially when you try to drag the map or zoom in and out. 

I'd start by asking you to play around with this page and compare the implementations.
https://deploy-preview-339--nextjs-knack-signs-markings.netlify.app/benchmark
_we can delete this test page and the utils before we merge (or in a follow-up PR)_

Screenshot 2026-02-04 at 6 21 23 PM

After seeing the difference, especially at 5,000+ marker counts, I moved forward to implement the GeoJSON layer. This PR should have all the same functionality as https://github.com/cityofaustin/atd-knack-signs-markings/pull/338.

Test this PR by moving around the map to load more signs, test the pop-up behavior, and try to break the things.
https://deploy-preview-339--nextjs-knack-signs-markings.netlify.app/

~One area that will require more testing is embedding this version of the app in the Knack test environment to ensure the sign data sourced from Knack is still behaving properly as `<Marker/>` components, including the distinct color. I'm going to request @chiaberry help me coordinate this.~

Update: See [comment](https://github.com/cityofaustin/atd-knack-signs-markings/pull/339#issuecomment-3855342332) below, Chia helped me test in Knack 🙏 

You can test the embed in the "test-30-may-2024-signs-and-markings-operations" Knack instance by logging in using the GIS QA credentials and going to the "Signs | Work Orders" view and filtering by "Issued". [Here's the URL](https://atd.knack.com/test-30-may-2024-signs-and-markings-operations#work-order-signs/?view_2565_filters=%5B%7B%22text%22%3A%22Issued%22%2C%22field%22%3A%22field_3265%22%2C%22value%22%3A%22ISSUED%22%2C%22operator%22%3A%22is%22%7D%5D&view_2565_page=1) that should land you there after login.

- Click through some Details pages to find Issued Work orders that have locations (here's an [example details page](https://atd.knack.com/test-30-may-2024-signs-and-markings-operations#work-order-signs/view-work-orders-details-sign/664b674284dc300029e338aa/)). 
- Scroll down below the Locations list to see the map. You'll have to zoom in a ways to see the yellow dots pulled from AGOL. They are zoom-gated to only render once you get close to a block-level zoom.
- Confirm the location(s) sourced from Knack render as blue markers. 
- Click on the Blue Markers and Yellow dots to see their pop-up behavior.

One implication of this change is we could reconsider our zoom gate on AGOL signs (currently 16.75) to show more assets without requiring the user to be as zoomed in.

### DIFF
--- a/signs-work-order-map/app/benchmark/page.tsx
+++ b/signs-work-order-map/app/benchmark/page.tsx
@@ -1,0 +1,270 @@
+"use client";
+
+import { useState, useMemo, useRef, useCallback, useEffect } from "react";
+import MapGL, {
+  MapRef,
+  Marker,
+  NavigationControl,
+  Source,
+  Layer,
+  MapMouseEvent,
+} from "react-map-gl/mapbox";
+import type { CircleLayerSpecification } from "mapbox-gl";
+import {
+  generateMockSigns,
+  generateMockGeoJSON,
+  useMarkerPerformanceBenchmark,
+  useLayerPerformanceBenchmark,
+  setupBenchmarkConsoleUtils,
+} from "@/utils/benchmark";
+import {
+  DEFAULT_MAP_PARAMS,
+  DEFAULT_MAP_PAN_ZOOM,
+} from "@/config/map";
+
+// Initialize console utils on load
+if (typeof window !== "undefined") {
+  setupBenchmarkConsoleUtils();
+}
+
+type RenderMode = "markers" | "layer";
+
+const MARKER_COUNTS = [100, 500, 1000, 2000, 5000, 10000];
+
+// Layer style for AGOL signs (matching current marker appearance)
+const BENCHMARK_SOURCE_ID = "agol-signs";
+
+const agolSignsLayerStyle: CircleLayerSpecification = {
+  id: "agol-signs-layer",
+  type: "circle",
+  source: BENCHMARK_SOURCE_ID,
+  paint: {
+    "circle-radius": ["interpolate", ["linear"], ["zoom"], 14, 4, 18, 8],
+    "circle-color": "#FFD700", // Gold/Yellow
+    "circle-stroke-color": "#000000",
+    "circle-stroke-width": 1,
+  },
+};
+
+export default function BenchmarkPage() {
+  const mapRef = useRef<MapRef>(null);
+  const [renderMode, setRenderMode] = useState<RenderMode>("markers");
+  const [markerCount, setMarkerCount] = useState(500);
+  const [isRunning, setIsRunning] = useState(false);
+  const [currentCount, setCurrentCount] = useState(0);
+
+  // Generate mock data based on current count
+  const mockSigns = useMemo(() => {
+    if (renderMode === "markers" && currentCount > 0) {
+      return generateMockSigns(currentCount);
+    }
+    return [];
+  }, [currentCount, renderMode]);
+
+  const mockGeoJSON = useMemo(() => {
+    if (renderMode === "layer" && currentCount > 0) {
+      return generateMockGeoJSON(currentCount);
+    }
+    return { type: "FeatureCollection" as const, features: [] };
+  }, [currentCount, renderMode]);
+
+  // Benchmarking hooks
+  useMarkerPerformanceBenchmark(
+    renderMode === "markers" ? mockSigns.length : 0,
+    isRunning
+  );
+
+  useLayerPerformanceBenchmark(
+    renderMode === "layer" ? mockGeoJSON.features.length : 0,
+    "agol-signs-layer",
+    isRunning
+  );
+
+  const runBenchmark = useCallback(() => {
+    // Reset state
+    setCurrentCount(0);
+    setIsRunning(true);
+
+    // Small delay to ensure reset, then set count
+    setTimeout(() => {
+      setCurrentCount(markerCount);
+    }, 100);
+
+    // Stop after benchmark completes
+    setTimeout(() => {
+      setIsRunning(false);
+    }, 3000);
+  }, [markerCount]);
+
+  const clearMarkers = useCallback(() => {
+    setCurrentCount(0);
+    setIsRunning(false);
+  }, []);
+
+  // Marker rendering
+  const markers = useMemo(() => {
+    if (renderMode !== "markers") return null;
+
+    return mockSigns.map((sign) => (
+      <Marker
+        key={sign.id}
+        longitude={sign.lng}
+        latitude={sign.lat}
+      >
+        <div
+          style={{
+            width: 8,
+            height: 8,
+            borderRadius: "50%",
+            backgroundColor: "#FFD700",
+            border: "1px solid #000000",
+            cursor: "pointer",
+            transform: "translate(-50%, -50%)",
+          }}
+        />
+      </Marker>
+    ));
+  }, [mockSigns, renderMode]);
+
+  const handleLayerClick = useCallback((e: MapMouseEvent) => {
+    if (e.features && e.features.length > 0) {
+      console.log("Clicked feature:", e.features[0].properties);
+    }
+  }, []);
+
+  return (
+    <div style={{ height: "100vh", display: "flex", flexDirection: "column" }}>
+      {/* Controls */}
+      <div
+        style={{
+          padding: "16px",
+          backgroundColor: "#f5f5f5",
+          borderBottom: "1px solid #ddd",
+          display: "flex",
+          gap: "16px",
+          alignItems: "center",
+          flexWrap: "wrap",
+        }}
+      >
+        <div>
+          <label style={{ marginRight: "8px", fontWeight: "bold" }}>
+            Render Mode:
+          </label>
+          <select
+            value={renderMode}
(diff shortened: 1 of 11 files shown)


signs-work-order-map/components/Map.tsx
```diff
@@ -100,30 +105,42 @@ export default function Map({ signs, messageType, editLocation }: MapProps) {
...
-  // Combine Knack signs with AGOL signs (only include AGOL signs if zoomed in enough)
-  const allSigns = useMemo(() => {
-    // Mark Knack signs with source
-    const knackSigns = signs.map((sign) => ({
-      ...sign,
-      source: "knack" as const,
-    }));
-    // Only include AGOL signs when zoomed in enough for performance
-    const visibleAgolSigns = isZoomedInEnough ? agolSigns : [];
+  const knackSigns = useMemo(
```
mateoclarke, line 118: here we're reverting to only treat the sign data from Knack as `<Marker/>`s

signs-work-order-map/components/Map.tsx
```diff
@@ -100,30 +105,42 @@ export default function Map({ signs, messageType, editLocation }: MapProps) {
...
+    () => signs.map((sign) => ({ ...sign, source: "knack" as const })),
+    [signs]
+  );
+  const signPins = useCreateSignPins(knackSigns, setPopupInfo);
 
-    return [...knackSigns, ...visibleAgolSigns];
-  }, [signs, agolSigns, isZoomedInEnough]);
+  const handleMouseEnter = useCallback(() => {
```
mateoclarke, line 124: There was some weird behavior with the cursor so I added this to be explicit about that a user could click a yellow dot as they hover over it. Kinda irrelevant since our power users will be on iPad mini tablets. But its here.

mateoclarke:
I tested with @chiaberry and confirmed that the sign data sourced from Knack is still rendered in the map as blue markers 👍  

_Also updated the README docs with more details on how to update the deploy preview to test the embed in the Knack staging environment :)_

Screenshot 2026-02-05 at 11 05 57 AM

One thing to note, the behavior in the iframe embed is such that whenever there are multiple Knack locations, the map auto zooms to a level that shows all the locations. This means you usually need to zoom in before the yellow dots are visible. This is intended behavior, but I'd suggest we ease up the zoom gate for the yellow dots so that they are more commonly visible when the iframe-embedded map loads.

signs-work-order-map/app/benchmark/page.tsx
johnclary, line 1: It was extremely fun using this tool 🙌

signs-work-order-map/components/Map.tsx
johnclary: Definitely good to have—nice 👍

signs-work-order-map/config/map.ts
```diff
@@ -24,6 +25,26 @@ export const DEFAULT_MAP_PAN_ZOOM = {
...
+export const AGOL_SIGNS_LAYER_STYLE: CircleLayerSpecification = {
+  id: AGOL_SIGNS_LAYER_ID,
+  type: "circle",
+  source: AGOL_SIGNS_SOURCE_ID,
+  paint: {
+    "circle-radius": 8,
+    "circle-color": "#FFD700",
+    "circle-stroke-color": "#000000",
+    "circle-stroke-width": 1,
+  },
```
johnclary, line 61: Could be a nice touch to interpolate the radius and stroke based on zoom. This would allow a larger circle size when zoomed in close. Maybe even have a "stop" at the highest zoom level that makes the circles really big so that it's super easy to tap on the one you want.

[Here's](https://github.com/cityofaustin/vision-zero/blob/57265428d6ad54915be317514879edfbe3b43c41/editor/configs/locationCrashesTable.ts#L83-L105) an example of this in the VZE.

johnclary: Using a white stroke around the circles might add contrast between overlapping points and also the background imagery. The yellow seems ok for contrast. For max contrast, going radical with `#00FFFF` or `#FF00FF` may be an improvement. Maybe something for Christina to test on the iPad on a sunny day 😎 

Here's what the markers look like on the VZE with the white stroke:

Image

johnclary (approved):
Wow, looks great! So cool to see the very significant performance difference in marker approaches. Thanks for setting us up with the benchmarking tool to play around with. 

🚢 🚢 🚢 🚢

signs-work-order-map/app/benchmark/page.tsx
frankhereford, line 1: +1 to what John said. 

Some of our team members were asking about when/how one might use AI in the dev cycle. To that, this is such a good example of how to really take advantage of the AI tooling we have at our fingertips. It takes a mental shift to get used to being able to conjure something in a few minutes where it might have taken a whole afternoon before. 🥇

signs-work-order-map/config/map.ts
frankhereford, line 61: Agreed -- and if you end up keeping a simpler solution with a fixed size, I thought the smaller size from the old map was really on-point. Signs are plentiful, they will have a dense map of symbology until you're just right down on them, so a small marker makes sense to me. Esp if they are all the same symbol -- then you can really crush down that size. This advice may not hold up well on the iPad, though.

signs-work-order-map/toolbox/benchmark/benchmarkUtils.ts
frankhereford, line 1: This is super minor - pls ignore if you want.

I'm not sure if you want to keep this file / tool around in the repo in `main` -- I'm fine if you do, but if so, I think it should probably make it out of this `utils` folder and into something that designates its one-time-use nature. Maybe `toolbox` or something like that. 

Or alternatively, a big call-out in the header comment to go with a entry in the README.

frankhereford (approved):
# 🤖 💪 

This is great Mateo -- I'm not sure how you could make a more compelling case for the  improved technique. Not only that, it's a superb example of leveraging AI in the dev cycle. 

🚢🚢🚢

Charlie-Henry (approved):
Dang really cool way of testing this! So much faster using the geojson layer

Image

signs-work-order-map/config/map.ts
roseeichelmann: The radius interpolation would be really cool... blue dots with white stroke does also look like better contrast to me which i imagine could be really helpful when out in the field!

roseeichelmann: I recall christina asking for the bigger sign markers in the last PR bc she was having a hard time clicking on the smaller ones on the iPad

signs-work-order-map/utils/mapUtils.tsx
```diff
@@ -199,46 +164,61 @@ export const useAGOLSignAssets = (
...
     loading,
     error,
   };
 };
 
+/**
+ * Converts a GeoJSON point feature from the AGOL layer (e.g. from map click) into a Sign
+ * for use in the popup.
+ *
+ * @param feature Clicked feature with Point geometry and AGOL properties (OBJECTID_1, etc.)
```
roseeichelmann: i think "feature" is accidentally repeated twice here

roseeichelmann (approved):
this is super cool! great idea to convert to a Layer instead of the individual Markers. the speed was greatlyyy reduced and the benchmark test page was such a great way to allow us to test this.

i noticed in the last PR that the markers got bigger when you hover on them but that seems to be removed in this PR, was that intentional?

also, i wanna reiterate that i think it would be super helpful to have the sign messages separated in the popup by a sort of list group like John mentioned in his comment [here](https://github.com/cityofaustin/atd-knack-signs-markings/pull/338#discussion_r2775012880), i find it difficult to grok all the different sign messages in the popup and instead they can feel like one long run on sign message

signs-work-order-map/toolbox/benchmark/benchmarkUtils.ts
mateoclarke: Thanks Frank, I took your suggestion: https://github.com/cityofaustin/atd-knack-signs-markings/pull/339/changes/6f4e5843c09414cc81ddd06278bbb653adcd0cd4

signs-work-order-map/utils/mapUtils.tsx
mateoclarke: It was intentional, but you're right, it looks weird and is poorly worded. In my text editor, `feature` is highlighted because its the name of the param. So I updated the description.

Screenshot 2026-02-12 at 1 32 55 AM

https://github.com/cityofaustin/atd-knack-signs-markings/pull/339/changes/325e581c6c4e1e37603190a2cd2889ecec04bbbe

PR #341, opened 2026-03-05
Adds ability to add existing locations

### LINKED ISSUE #21665: Add the ability to add existing locations to the sign work order map

mateoclarke:
Closes https://github.com/cityofaustin/atd-data-tech/issues/21665

This PR adds the ability to add existing sign locations from the map (AGOL asset locations) in addition to creating new locations by placing a pin. In order to achieve this, we've used a design pattern from Moped where you toggle your map edit mode (pin drop vs select existing) using a map control in the top right corner.

Screenshot 2026-03-04 at 5 24 37 PM

## Testing

Use the [test Knack app `test-30-may-2024-signs-and-markings-operations`](https://atd.knack.com/test-30-may-2024-signs-and-markings-operations#work-order-signs/view-work-orders-details-sign/664b674284dc300029e338aa/) and GIS QA credentials (from 1Password).

1. Log in to the Knack test app with GIS QA credentials
2. Open an **Issued** Work Order 
    - ex: Signs | Work Orders → filter by Issued → open a Details page, then open a work order like:
https://atd.knack.com/test-30-may-2024-signs-and-markings-operations#work-order-signs/view-work-orders-details-sign/664b674284dc300029e338aa/
3. Scroll to the map below the Locations list. The map should be loading from this PR’s deploy preview.
4. Create Location (default)
    - The new toggle in the top right should show “Create Location” selected.
    - You should see the red pin at map center and “Add Location” button visible in the bottom-left.
    - Pan/zoom or use search to move the pin
    - Add a new location with “Add Location” button and confirm it appears in the Locations table above and on map as a new ~blue~ yellow marker.
5. Select Existing
    - Switch the toggle to “Select Existing”.
    - “Add Location” button should disappear
    - You may need to zoom in until blue AGOL dots appear.
    - Click a blue dot to trigger the popup with Location ID and sign assets listed.
    - Click “Add this Location” in the popup.
    - Confirm a new row is added to the Locations table with the correct location, correct `ASSET_LOCATION_ID` value, and that the map/iframe reloads in Create mode with “Add Location” visible again.
6. Switching back to Create Location
    - In Select Existing, switch the toggle back to “Create Location”.
    - The “Add Location” button should reappear in the bottom left.
    - Red center pin and normal “create by panning” behavior should work as before.


mateoclarke:
Based on DM's with @ChristinaTremel, I'm going to:
- [x] Confirm we are capturing the AGOL `Location ID` in the "work order signs locations table"
- [x] Update the Knack signs pins to `#FFC600`
- [x] Make sure the Knack test environment is prepped for user testing by end of day Monday 3/16

knack/iframeMapMessenger.js
```diff
@@ -47,10 +48,18 @@
...
+        id="mapIFrame" style="width: 100%;height: 523px;"></iframe>`,
       ).appendTo($viewSelector);
     }
 
+    // Always hide the ASSET_LOCATION_ID field — it is populated
+    // programmatically and never needs to be visible to the user
+    $("#kn-input-field_4461").closest(".kn-input").css({
+      visibility: "hidden",
+      height: 0,
+      overflow: "hidden",
```
mateoclarke, line 59: Perhaps this code belongs in the Knack css file, but I find it useful to include here with the comment next to the rest of the iframe messenger code. Especially if and when we need to update field ID numbers.

signs-work-order-map/app/page.tsx
```diff
@@ -1,22 +1,89 @@
...
+    <div className={styles.map} style={{ display: "flex", alignItems: "center", justifyContent: "center" }}>
+      <div className="map-status-indicator map-status-indicator--loading">
+        <span aria-hidden="true">⏳</span>
+        <span>Loading map...</span>
+      </div>
+    </div>
+  ),
 });
 
+function LoadingOrStandalonePrompt({
```
mateoclarke, line 29: The point of this code was to address a minor visual bug that was annoying me. Without the delays, the map would load in the default location (neat House Park on Lamar) and then once the Knack Locations loaded, the map would jump to the proper place. 

So what I'm doing here is waiting for the Knack payload to come through before rendering the map. And in case we run the app outside of the Knack iframe, providing an opt out to load the map anyways.

signs-work-order-map/components/CrosshairIcon.tsx
```diff
@@ -0,0 +1,24 @@
+interface CrosshairIconProps {
+  size?: number;
+}
+
+export default function CrosshairIcon({ size = 14 }: CrosshairIconProps) {
```
mateoclarke: I don't think we're using an icon library and Cursor did a pretty nice job making these SVG icons as components. Hope thats an ok path forward. If we wanted to adopt a library, this [Lucide library](https://lucide.dev/icons/crosshair) seems pretty slick

signs-work-order-map/components/Map.tsx
```diff
@@ -136,12 +160,34 @@ export default function Map({ signs, messageType, editLocation }: MapProps) {
...
+      if (feature) {
+        // Log the raw AGOL feature data for debugging/inspection
+        // Includes geometry and all available properties for the clicked pin
+        // eslint-disable-next-line no-console
+        console.log("AGOL sign feature clicked:", feature);
+      }
+
+      if (sign) {
+        // eslint-disable-next-line no-console
+        console.log("Derived AGOL sign object:", sign);
```
mateoclarke: We might want to remove this console.log lines 171 and 176 before we take this app to production but for troubleshooting, it was helpful for me to have them so I'm going to leave them for now.

Charlie-Henry (approved):
I was able to follow your testing instructions and everything was working good for me 👍 

I'm not going to be very helpful with this comment but I feel like this "Add location" button seems a little off to me, I guess it doesn't really feel like a button to me? Maybe it's the hover over which seems to only change the blue to a slightly darker blue?

Image

ChristinaTremel:
I went out to the field with Ted (one of our sign techs) and tested the work order map in downtown and north Austin. I was really happy with the performance of the map. 👏🏼 The map, basemap and AGOL locations were loading almost immediately and the process of adding the points/reloading the page took just a few seconds (faster than adding attachments in Knack 🫠). 

We noticed a couple of things while in the field we were hoping could be modified: 

### Toggling between "Create Location" and "Select Existing", 
Every time a point was added from the "Select Existing" toggle, there would be a quick reload and the map defaults back to the "Create Location" toggle. Many times if the techs are doing maintenance on existing signs, they are maintaining multiple locations along the street. If they are installing new signs, they are installing several new locations. 

Is it possible to keep the toggle option they had selected after reloading the map? So if they were on the "Select Existing", when it reloads the map stays on the "Select Existing" toggle? 

Video for reference
https://github.com/user-attachments/assets/93e97a45-99e0-4cb9-89ba-e7f0962a7905

### Pop-ups cutting off on the map
Some of the pop-ups have 4-5 sign messages. It doesn't always happen, but sometimes when opening the pop-up, the pop-up cuts off on the map. If we chose "Select Existing", the button to add the location to the map isn't visible on the screen to select. The workaround is just moving the map a bit to see the button, but is it possible to restrict the pop-ups from cutting off on the map? 

Video for reference
https://github.com/user-attachments/assets/24f0c33d-1985-4189-b8b4-43b4b2207841

knack/iframeMapMessenger.js
```diff
@@ -7,9 +7,10 @@
 (function () {
   var myView = window.viewIdsArray.shift(0);
 
-  const nextAppUrl =
-    "https://deploy-preview-339--nextjs-knack-signs-markings.netlify.app";
-  // const nextAppUrl = "http://localhost:3000";
+  console.log("mateo is testing: ", new Date().toISOString());
```
johnclary: 👀

signs-work-order-map/app/page.tsx
johnclary: Really nice. Some minor feedback: Instead of "No knack payload found", replace it with something more user friendly. "No work order locations found"? Or whatever the parlance is that techs will understand. 

Image

By the way, I hit this prompt because my browser location is outside the bounds of the map. Once I closed the modal I was able to zoom in on Austin to see what I needed to see.

johnclary: One more more minor thing—would you mind adding a docstring above this component that mentions what it does? "Shows a dismissible modal when knack payload fails to load..."

signs-work-order-map/components/CrosshairIcon.tsx
johnclary: We're using Lucide in VZE now as well 💘 

The crosshair icon initially threw me off for a sec because I associated it with the location services button that is on the left side of the map. Maybe [`pointer`](https://lucide.dev/icons/pointer) or [`square-mouse-pointer`](https://lucide.dev/icons/square-mouse-pointer) would be better?

Image

signs-work-order-map/components/Map.tsx
```diff
@@ -47,6 +52,17 @@ import {
...
   const [popupInfo, setPopupInfo] = useState<Sign | null>(null);
+  const [locationMode, setLocationMode] = useState<LocationMode>("create");
+
+  const handleLocationModeChange = useCallback((mode: LocationMode) => {
+    setLocationMode(mode);
+    sendLocationModeToParent(mode);
+  }, []);
+
+  const showLocationToggle =
+    messageType === "WORK_ORDER_SIGNS" || messageType === "EDIT_LOCATION";
```
johnclary: Not related to this PR, but I wonder if we could update the message types to be a little more clear about what's happening in the Knack app. If these names could resemble events, then for example `"WORK_ORDER_SIGNS"`  becomes `"LOAD_WORK_ORDER_DETAILS_PAGE"`.

signs-work-order-map/components/Map.tsx
```diff
@@ -58,7 +74,11 @@ export default function Map({ signs, messageType, editLocation }: MapProps) {
     west: number;
   } | null>(null);
   const [mapLoaded, setMapLoaded] = useState(false);
+  const [initialViewApplied, setInitialViewApplied] = useState(false);
   const [zoom, setZoom] = useState<number>(DEFAULT_MAP_PAN_ZOOM.zoom);
+  const locationModeRef = useRef<LocationMode>(locationMode);
+  locationModeRef.current = locationMode;
```
johnclary, line 93: thanks for another example of how to bring `refs` into the mix with reactivity 🙏

signs-work-order-map/components/Map.tsx
```diff
@@ -243,26 +296,48 @@ export default function Map({ signs, messageType, editLocation }: MapProps) {
...
         setMapLoaded(true);
         setZoom(initialZoom);
         updateMapBounds();
-        sendLatLonToParent(mapLatLon);
+
+        if (signLocationBounds) {
+          map.fitBounds(signLocationBounds, {
+            padding: 100,
+            maxZoom: 16,
+            duration: 0,
```
johnclary: These are jumping out to me as settings that should live in `@configs/map`

signs-work-order-map/components/Map.tsx
```diff
@@ -273,24 +348,50 @@ export default function Map({ signs, messageType, editLocation }: MapProps) {
...
+          }}
+        />
       )}
       {isZoomedInEnough && agolError && (
         <MapStatusIndicator
           type="error"
           message={`Error loading signs: ${agolError}`}
+          position={{
+            top: showLocationToggle ? "48px" : "10px",
+            right: "10px",
```
johnclary: noting that all of these `px` values are repeated quite a few times..

signs-work-order-map/components/PinIcon.tsx
```diff
@@ -0,0 +1,21 @@
+interface PinIconProps {
+  size?: number;
+}
+
+export default function PinIcon({ size = 14 }: PinIconProps) {
```
johnclary: It'd be nice to have a comment that leaves a breadcrumb of where these are sourced from.

signs-work-order-map/utils/iFrameMessenger.tsx
```diff
@@ -31,3 +31,32 @@ export function sendLatLonToParent(coords: LatLon) {
...
+) {
+  // eslint-disable-next-line no-console
+  console.log("Sending EXISTING_LOCATION_SELECTED message", {
+    lat: coords.latitude,
+    lng: coords.longitude,
+    assetLocationId,
+  });
+  window.parent.postMessage(
+    {
+      message: "EXISTING_LOCATION_SELECTED",
```
johnclary, line 47: Yep, this message name makes it super clear what's happening in the app 👍

signs-work-order-map/utils/mapUtils.tsx
```diff
@@ -118,7 +118,7 @@ export const useCreateSignPins = (
               key={`marker-${sign.id}`}
               longitude={sign.lng}
               latitude={sign.lat}
-              color={sign.isLocationDetailPage ? "red" : undefined}
+              color={sign.isLocationDetailPage ? "red" : "#FFC600"}
```
johnclary, line 121: I know there's refinement left do around the various colors and markers, but wanted to note that when you click one of the gold markers, the popup hides the marker entirely, making it look like the popup is attached to the underlying existing location.

Image

Image

signs-work-order-map/components/SignPopup.tsx
```diff
@@ -86,6 +92,18 @@ export default function SignPopup({ popupInfo, setPopupInfo }: SignPopupProps) {
...
+            {locationMode === "select_existing" &&
+              onSelectExistingLocation && (
+                <button
+                  className="btn btn-primary btn-sm w-100 mt-2"
+                  onClick={() => {
+                    onSelectExistingLocation(popupInfo);
+                    setPopupInfo(null);
+                  }}
+                >
+                  Add this Location
```
johnclary, line 101: It feels like there is missing user feedback whenever you add a location to the work order (when creating a new one or adding an existing).

Would it make sense to navigate to the location details page after adding a location?  Or is the normal workflow that the technician adds all the locations, then goes to the location details for each one and adds signs? 

Alternatively, maybe in a follow-up issue we could add a dismissible message to the map that says "Sign location added ✅ ".

signs-work-order-map/components/LocationModeToggle.tsx
```diff
@@ -0,0 +1,40 @@
...
+export default function LocationModeToggle({
+  mode,
+  onModeChange,
+}: LocationModeToggleProps) {
+  return (
+    <div className="location-mode-toggle">
+      <button
+        className={`location-mode-toggle__btn ${
+          mode === "create" ? "location-mode-toggle__btn--active" : ""
+        }`}
```
johnclary, line 24: I've mentioned this before, and it might be a minority opinion, but still throws me off that when the map initially loads it is already in "Create Location" mode. 

Just brainstorming here—what do y'all think about this flow:

- On initial map load, show the **Add location**  in the top right corner.

Image

- Clicking this button reveals the new toggle with **Create location** active by default. The red marker appears at the center of the map, and now it has an always-open popup with a button that says "Add this location"


Image

- The user can alternatively switch to **Select location** mode, exactly as implemented here.


I'm also having doubts about whether the "Select location" toggle is needed, or if you could always show the **Add the location** in the pop-ups 🤔 

Image

signs-work-order-map/components/SignPopup.tsx
```diff
@@ -36,7 +43,6 @@ export default function SignPopup({ popupInfo, setPopupInfo }: SignPopupProps) {
 
   return (
     <Popup
-      anchor="top"
```
johnclary, line 39: I noticed this after I left a comment elsewhere about the marker being obscured by its popup. Could be related?

signs-work-order-map/utils/mapUtils.tsx
johnclary: Again, I know there's ongoing refinement here, here's an example of unifying the marker and popup styles by adding a `New` badge in side the popup for new locations, and styling locations on the current work order in magenta.

Image

Image

johnclary (approved):
Looking great! Test steps all worked perfectly. I hope my feedback is helpful. Thanks for all your work on this. Very cool to see the existing locations functionality fall into place!! 🚢

signs-work-order-map/components/PinIcon.tsx
mateoclarke: Decided to go ahead and just `npm install lucide-react`

signs-work-order-map/components/SignPopup.tsx
mateoclarke: Yeah good call out... I'm not sure why I deleted that line at all.

mateoclarke:
Thanks for the feedback @johnclary, I'm going to run your unaddressed suggestions by @ChristinaTremel and propose that if we take them on, we do it in the next PR:
- [Alternative "Add Location" flow starting without red pin](https://github.com/cityofaustin/atd-knack-signs-markings/pull/341#discussion_r2975906456)
- [Give user feedback when Location adds successfully](https://github.com/cityofaustin/atd-knack-signs-markings/pull/341#discussion_r2975850471)
- [Style current work order locations in another color and unify markers](https://github.com/cityofaustin/atd-knack-signs-markings/pull/341#discussion_r2976081001)

signs-work-order-map/components/SignPopup.tsx
mateoclarke: Actually, I did need this line removed so that the popups stay inside the viewport. I'm going to revert it to how it was and live with the popup hiding the marker as an issue to resolve down the line.

mateoclarke: Screenshot 2026-04-01 at 10 31 02 AM

signs-work-order-map/components/LocationModeToggle.tsx
roseeichelmann: I really like the workflow you laid out John, and I also found the Select existing toggle to be confusing. To me it was an odd workflow that you have to toggle  "Select existing" to get the "Add this Location" button to show in the existing location popups. I think it would be simpler and more intuitive if the button is always available like john mentioned rather than needing to toggle anything, and instead of the toggle its just a button to go into Location add mode which shows the red pin.

roseeichelmann (approved):
Test steps all worked as expected! I liked some of johns ideas for the workflow but it makes sense to work those potential changes into a future PR.

Can the text for Asset location ID in this sc be made white bc its almost invisible rn:

Image

mateoclarke:
Hey @roseeichelmann, I'm a little confused by your screen grab. That Asset Location ID field is supposed to be hidden from the user. Are you always seeing it? The only way it should render is if you tinker with the CSS classes/rules like I've done as an example in my screenshot. Could you maybe do a video recording or elaborate on what you're seeing?

Screenshot 2026-04-02 at 2 27 27 PM

ChristinaTremel (approved):
Hey Mateo! I tested out the map and you've addressed all the issues I was experiencing previously. 🙂 Looking at other feedback I'm super interested in [giving user feedback when Location adds successfully](https://github.com/cityofaustin/atd-knack-signs-markings/pull/341#discussion_r2975850471) and [unify markers](https://github.com/cityofaustin/atd-knack-signs-markings/pull/341#discussion_r2976081001), but as discussed I like the idea of addressing those in a new issue!

I also didn't see the Asset Location ID visible on the map in the bottom left corner when I tested and don't want the users the see that - so everything looks good to me!

PR #343, opened 2026-04-20
Configures pop-ups for sign locations added to work order

mateoclarke:
## Associated issue
Closes https://github.com/cityofaustin/atd-data-tech/issues/26881

## Summary

Two related pieces of work shipped together on this branch:

### 1. Pop-up enhancements for sign locations added to a work order
- Popups now prefer **Location ID** (from AGOL `ASSET_LOCATION_ID` / Knack `field_4461`) and fall back to **Spatial ID** only when no Location ID is present.
- Lat/lng is no longer shown in the popup.
- When a Knack sign-location pin shares a lat/lng with an AGOL sign (within ~5 m), the AGOL marker is hidden and its attributes (including sign messages) are merged into the Knack pin's popup.
- The ID row in the popup is now a link that opens the Location Details page in Knack.

### 2. Symbology + default mode + "Add Location" button styling
- Knack sign locations now render as circular points (not pins) with state-based styling:
  - **Yellow** — existing Knack sign location
  - **Green with a "NEW" label** — newly-created location (no AGOL `ASSET_LOCATION_ID`)
  - **Red** — the location currently being viewed on a Location Details page
- Map defaults to **Select Existing** mode on first load; user's last choice is still persisted in `sessionStorage`.
- The "Add Location" button in Create Location mode is restyled to match the popup's Bootstrap `btn-primary` and is centered at the bottom of the map.

Screenshot 2026-04-22 at 4 24 38 PM

Screenshot 2026-04-22 at 4 24 59 PM

## Testing

Use the [test Knack app `test-30-may-2024-signs-and-markings-operations`](https://atd.knack.com/test-30-may-2024-signs-and-markings-operations#work-order-signs/view-work-orders-details-sign/664b674284dc300029e338aa/) and GIS QA credentials (from 1Password).

**Steps to test:**

### 1. Default mode is "Select Existing"
1. If necessary, clear `sessionStorage` for the Knack app origin (DevTools > Application > Storage > Clear site data), then load the Add Sign Location view on an existing work order.
1. Expected: The **Select Existing** toggle is active by default (not Create Location).
1. Switch to **Create Location**, reload the page.
1. Expected: The map restores **Create Location** — the last-used mode is still persisted across reloads.

### 2. Pop-ups prefer Location ID and link to details & No lat/lng in popups
1. In **Select Existing** mode, click an AGOL sign that is **not** already added to the work order.
1. Expected popup shows: sign message(s), no lat/lng row, no Spatial ID.
[...]

PR #344, opened 2026-04-22
Fixes jumping page issue when adding a location point and adds fullscreen mode

### LINKED ISSUE #28011: Explore fixes to the jumping page issue when adding a location point
### Description

- Explore if it’s possible to fix the jumping of the page after adding a location point
   - If that’s possible, add a banner confirming that the sign location has been added
- Explore if it's possible to expand the map to full page to fix that jumping page issue

Also:

> Upon initial testing the only thing I'm noticing is when a location is added, the map zooms out further than I would expect it to and I have to zoom back in every time. I think if it could zoom out less when the map reloads that would be helpful!

mateoclarke:
Closes https://github.com/cityofaustin/atd-data-tech/issues/28011

## Summary
This branch (_somewhat_) addresses the map reload and page jump problem on the Work Order Details page, adds a fullscreen map toggle, and introduces confirmation banners.

1. **Persistent iframe** (no more map reloads)
    - The map iframe is now placed in a `#mapIFrameContainer` div **before** the Knack view (`view_2573`) as a DOM sibling, rather than inside the view at the bottom. This means that when Knack re-renders the view after a form submission, the iframe survives intact and the view that reloads doesn't affect the map (as much as before).
    - On re-render, the script detects the existing iframe and pushes fresh data via `postMessage` instead of creating a new one.
2. **Scroll & layout stabilization** (no more page jumps, kinda)
    - Scroll position is saved before form submission (`window.__mapScrollY`) and restored in multiple passes after the DOM rebuilds. This means the page still jumps (can't do much about how Knack handles this), but once Knack is done reloading the view, its scrolls you back where you were. 
3. **Fullscreen map toggle**
    - A fullscreen button (positioned below the Create/Select toggle) expands the map to fill the viewport. Esc or clicking the button exits fullscreen.
4. **Location confirmation banner** (React-side)
    - When a location is added or removed, the Knack script sends a `SHOW_BANNER` `postMessage` to the iframe.
A new `useBannerMessage` hook in the React app listens for it, and a MapBanner component renders a green banner that auto-fades after 5 seconds and is dismissable.

## Testing

I created a new Knack test instance so I could do development on this branch locally while #343 is still being reviewed and worked on. It has been set now to use this deploy preview branch. 

Test environment: [**TEST | 30 APRIL 2026 | Signs and Markings Operations**](https://atd.knack.com/test--30-april-2026--signs-and-markings-operations#work-order-signs/view-work-orders-details-sign/664b85725d3e620028be63c0/)

_Log in using "Test GIS - Signs & Markings Tracker" credentials in 1pw._

1. Go to an issued work orders details page, [example](https://atd.knack.com/test--30-april-2026--signs-and-markings-operations#work-order-signs/view-work-orders-details-sign/664b674284dc300029e338aa/)

**Persistent iframe & scroll:**
1. Navigate to a work order with existing locations. Verify the map loads with sign pins.
2. Scroll down so the map is partially out of view. Add a location via "Create Location" > "Add this Location". Confirm the map does not fully reload (no tile flicker) and the page does not jump to a new spot (once the transaction completes, the scroll should be where it was).
3. Do the same with "Select Existing" > click a sign > "Select this Location". 

**Fullscreen:**
1. Click the fullscreen button (below the Create/Select toggle, top-right). Confirm the map fills the viewport.
2.  Press Esc or click the button again. Confirm it exits fullscreen.
3. Test the zoom in and out buttons. (@ChristinaTremel report having issues but I couldn't reproduce)
4. In fullscreen again, try to add locations and confirm all works similar to the smaller iframe mode.

**Mode switch & popup:**
1. In "Select Existing" mode, click a sign to open its popup. Toggle to "Create Location". Confirm the popup closes automatically.

**Banner:**
1. Add a location (either mode). After the table refreshes, confirm a green "Location added" banner appears on the map.
2. Wait 5 seconds — confirm the banner fades out.
3. Add another location and click the X on the banner before 5 seconds — confirm it dismisses immediately.


ChristinaTremel:
I'm a fan of the fullscreen mode and the jumping page fix is definitely feeling better! 🙌🏼  The only thing that immediately jumped out to me as a challenge is it seems like I'm not able to use the zoom in `+` in the bottom right corner of the map! It doesn't work in the standard mode or full screen mode. I'll test on an iPad mini tomorrow :)

knack/iframeMapMessenger.js
```diff
@@ -39,16 +42,124 @@
...
+      .off("keydown.mapFullscreen")
+      .on("keydown.mapFullscreen", function (e) {
+        if (
+          (e.key === "Escape" || e.keyCode === 27) &&
+          $("body").hasClass("map-iframe-fullscreen")
+        ) {
+          exitFullscreen();
+        }
+      });
+  }
```
frankhereford, line 113: I find that if i scroll the map after I go into full-screen mode, I can no longer use the ESC button to return back into normal mode. I tested this in Arc and Chrome.

signs-work-order-map/components/MapBanner.tsx
frankhereford, line 1: This is a real nice touch, I like the toast like notification.

frankhereford (approved):
This works really well. I found a small unexpected behavior with the ESC key, but I think this is ready to ship anytime. Thank you for this refinement on this good looking app!

knack/iframeMapMessenger.js
```diff
@@ -2,6 +2,9 @@
  * This code is imported into the Knack app and executed as custom vanilla JS that interacts with the NextJS app
  * via iFrame messaging.
  *
+ * The iframe is placed in a container BEFORE the Knack view (as a sibling, not
+ * a child) so that when Knack re-renders the view (e.g. after adding a
+ * location), the map iframe is outside the blast radius and survives intact.
```
chiaberry, line 7: such an upgrade!

knack/iframeMapMessenger.js
chiaberry: This happened to me too, and hitting escape made the button to return to normal mode disappear, so I was stuck in fullscreen mode

chiaberry (commented):
Moving the iframe out of the view is such a smart thing to do, I am glad you did it! This is out of the scope of this PR, but when I was testing I noticed that while on the location details view, the agol markers are visible and have the button to "add this location" on their popup, but when I click the button, nothing is added. The console log says the message is received, but I dont think knack has the form needed to receive the input on that view. (That is me going off of hazy memory)

I tested on an old ipad mini and I was able to zoom in and out using the buttons and touch input. 

I had the same experience as Frank that when I went into fullscreen mode, sometimes I couldnt get out of it with the escape key. And then the icon the go back to normal mode disappeared so i had to refresh the page.

knack/iframeMapMessenger.js
johnclary: Interesting—i did not have this problem fwiw

knack/iframeMapMessenger.js
```diff
@@ -39,16 +42,124 @@
...
 
+  /**
+   * Adds a "Fullscreen" button that expands the map iframe to fill the browser
+   * viewport via CSS (not the native Fullscreen API, which requires extra
+   * permissions for cross-origin iframes). Pressing Esc exits fullscreen.
+   *
+   * Styles live in knack/index.css (`.map-fullscreen-btn`,
+   * `body.map-iframe-fullscreen`, `.map-iframe-fullscreen__frame`).
+   */
+  function setupMapFullscreenToggle($iframe) {
```
johnclary, line 53: all of this is fun to read—a cool solution!

A few minor issues:
- I am getting a right scrollbar in fullscreen I assume because of the "not native" aspect of it, and the scrollbar is covering up the +/- controls slightly. I wonder if applying an overflow hidden to rest of the document while in fullscreen is an option. 
- When you zoom in/pan, the "Loading Sign Assets" message appears, but it appears underneath the fullscreen button. Ditto for the "Zoom in to view sign assets message." Maybe those messages would be better placed in top center, where the location confirmation message appears.

Image

knack/iframeMapMessenger.js
```diff
@@ -39,16 +42,124 @@
...
+      var formPoll = setInterval(function () {
+        if (moveLatLonForm()) clearInterval(formPoll);
+      }, 100);
+      setTimeout(function () { clearInterval(formPoll); }, 10000);
+    }
+
+    // Stabilize layout during Knack table re-renders: give the view a
+    // min-height so the page height doesn't collapse when Knack empties
+    // and repopulates the table DOM.
+    var $view = $(myView);
```
johnclary, line 165: I'm not quite following what effect this should be having. There is a page jump when some Knack UI elements disappear but then the map scroll recovers. Which is a big improvement over the before state but just want to mention that there is still some jumping.

knack/iframeMapMessenger.js
```diff
@@ -59,69 +170,128 @@
       overflow: "hidden",
     });
 
+    // Restore scroll position that was saved before form submission.
+    // Delayed slightly so it runs after Knack finishes rebuilding the DOM.
+    if (window.__mapScrollY != null) {
```
johnclary, line 186: Nice—this works really well 🚀

knack/iframeMapMessenger.js
```diff
@@ -59,69 +170,128 @@
...
+      var actionMsg = window.__mapLocationAction;
+      var bannerVariant = actionMsg.indexOf("removed") !== -1 ? "error" : "success";
+      delete window.__mapLocationAction;
+      setTimeout(function () {
+        sendMessageToApp(
+          { message: "SHOW_BANNER", text: actionMsg, variant: bannerVariant },
+          $iframe[0].contentWindow,
+        );
+      }, 300);
+    }
```
johnclary, line 210: Fancy!

signs-work-order-map/components/Map.tsx
```diff
@@ -265,22 +270,63 @@ export default function Map({ signs, messageType, editLocation }: MapProps) {
   }, [initialCenter]);
 
   /**
-   * Zoom to bounding box containing sign location pins
-   * and set "add location marker" coordinates to center
+   * When sign data changes: initial / delete / mixed → fit bounds.
+   * Pure addition (new row(s), no removals) → keep current zoom and center on the new pin.
+   * Identical sign set (Knack re-sent same records) → no camera change.
```
johnclary, line 275: Orchestrating this map state recovery is so cool 🙌

signs-work-order-map/utils/iFrameMessenger.tsx
```diff
@@ -58,6 +58,42 @@ export function useIFrameMessenger() {
   return message;
 }
 
+export type BannerVariant = "success" | "error";
+
+/**
+ * Listens for SHOW_BANNER postMessages from the Knack parent window.
+ * Returns the banner text, variant, and a function to clear it.
+ */
+export function useBannerMessage() {
```
johnclary, line 67: I am lazily not looking into what it would take to show the AGOL-related messages using this function instead of rendering them the way they are now—but that could be a nice touch

johnclary (approved):
Wow—so many upgrades in this PR. Seeing all the moving parts working is really impressive. This is a huge step forward 🙌  🚢

knack/iframeMapMessenger.js
roseeichelmann: I had the same issue as Chia & Frank. Sometimes ESC would work, but if i scrolled or moved the map it wouldnt work anymore and sometimes the button would disappear like chia said

knack/iframeMapMessenger.js
roseeichelmann: yeah this is so nice!

roseeichelmann (commented):
Tested on firefox,  i really love the new banner update! seeing that confirmation is so helpful.

I experienced what Christina said, the zoom in/out buttons were not working for me a lot of the time on both fullscreen and minimized, so i think there is an issue outside of the scroll bar covering the buttons on fullscreen (like john mentioned). The mouse wouldnt become clickable when hovering the buttons.

I also noticed what chia said about the button for adding location being visible on the location details page map, but it doesnt do anything.

Im still experiencing jumping when I click the Add location button, but as John said if you wait the scroll will recover.

mateoclarke:
@frankhereford @chiaberry @roseeichelmann, would y'all mind trying to reproduce the esc button now? I think I resolved it. Escape was only handled on the Knack parent page, so once you click or pan the map, focus moves into the cross-origin iframe and the parent never sees the keypress. The React app now forwards Escape to Knack via postMessage, and Knack exits fullscreen when it receives that message. 😅

@ChristinaTremel, also zoom in/zoom out should be better now too.

roseeichelmann (approved):
Mateo, the problems with the ESC key and the zoom in/out buttons have been resolved, this looks great to me!! Thanks for working through this 🚀

frankhereford:
I got to check this out again, and the `esc` key is working great. Thanks for fixing!

chiaberry (approved):
I can escape now! 🚪

mateoclarke:
@ChristinaTremel, I'd like to close this one out but still need your approval. Thanks!

ChristinaTremel (approved):
Apologies for the delay in approval! I was able to test in late July and the buttons/full screen mode were working so nicely. ✨

PR #345, opened 2026-05-19
Match by location instead of proximity

mateoclarke:
## Associated issue

Addresses this feedback from John:
https://github.com/cityofaustin/atd-knack-signs-markings/pull/343#discussion_r3220790892

## Summary

This PR uses a better approach for sign deduplication and fixes a bug in the Select Existing flow.

**Match Knack and AGOL signs by Location ID (not distance)**
`enrichKnackSignsWithAgol` no longer pairs Knack work-order pins with nearby AGOL features using a 5 meter proximity threshold. It now joins on `ASSET_LOCATION_ID` (Knack field_4461 / AGOL ASSET_LOCATION_ID). That avoids incorrectly merging or hiding the wrong AGOL marker when many signs are close together.

**iframe messaging bug fix**
`useIFrameMessenger` only accepts inbound Knack message types and safely parses `event.data`. Outbound messages such as `EXISTING_LOCATION_SELECTED` and `LOCATION_MODE_CHANGE` are ignored so they cannot overwrite the last valid work-order payload. `useFormatSignsRecords` guards against missing `payload.records` so the map does not crash if an unexpected message shape is received.

## Testing

Use the [test Knack app `test-30-may-2024-signs-and-markings-operations`](https://atd.knack.com/test-30-may-2024-signs-and-markings-operations#work-order-signs/view-work-orders-details-sign/664b674284dc300029e338aa/) and "Test GIS - Signs & Markings Tracker" credentials (from 1Password).

**Steps to test:**

1. Open a work order detail page and toggle to "Select Existing" in the map view
1. Zoom in until blue AGOL dots appear.
1. Click a blue dot to open the popup, click "Add this Location".
1. Map data will refresh, zoom back in to where you were (I know this part is annoying, trying to address it in #344 
Expected: No runtime error; form submits; map stays usable until Knack sends updated data; new yellow pin appears after reload/data push and appears to merge data where the blue dot was.

### DIFF
--- a/signs-work-order-map/utils/iFrameMessenger.tsx
+++ b/signs-work-order-map/utils/iFrameMessenger.tsx
@@ -2,22 +2,56 @@
 import { useEffect, useState } from "react";
 import { KnackToIFrameMessage, LatLon, LocationMode } from "@/types/map";
 
+const KNACK_TO_IFRAME_MESSAGE_TYPES = [
+  "LOAD_WORK_ORDER_LOCATION_DETAILS_PAGE",
+  "LOAD_WORK_ORDER_DETAILS_PAGE",
+  "OPEN_LOCATION_EDITOR",
+] as const;
+
+function parseMessageEventData(data: unknown): unknown {
+  if (data == null) return null;
+  if (typeof data === "string") {
+    try {
+      return JSON.parse(data);
+    } catch {
+      return null;
+    }
+  }
+  if (typeof data === "object") return data;
+  return null;
+}
+
+function isKnackToIFrameMessage(value: unknown): value is KnackToIFrameMessage {
+  if (!value || typeof value !== "object" || !("message" in value)) {
+    return false;
+  }
+  const { message } = value as { message: unknown };
+  return (
+    typeof message === "string" &&
+    (KNACK_TO_IFRAME_MESSAGE_TYPES as readonly string[]).includes(message)
+  );
+}
+
 export function useIFrameMessenger() {
   const [message, setMessage] = useState<KnackToIFrameMessage | null>(null);
 
   useEffect(() => {
-    const consoleMessage = (event: MessageEvent) => {
+    const onMessage = (event: MessageEvent) => {
       if (event.origin !== "https://atd.knack.com") {
         return;
       }
 
-      const data: KnackToIFrameMessage = JSON.parse(event?.data);
-      setMessage(data);
+      const parsed = parseMessageEventData(event.data);
+      if (!isKnackToIFrameMessage(parsed)) {
+        return;
+      }
+
+      setMessage(parsed);
     };
-    window.addEventListener("message", consoleMessage);
+    window.addEventListener("message", onMessage);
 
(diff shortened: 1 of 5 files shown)


signs-work-order-map/utils/iFrameMessenger.tsx
```diff
@@ -2,22 +2,56 @@
 import { useEffect, useState } from "react";
 import { KnackToIFrameMessage, LatLon, LocationMode } from "@/types/map";
 
+const KNACK_TO_IFRAME_MESSAGE_TYPES = [
+  "LOAD_WORK_ORDER_LOCATION_DETAILS_PAGE",
+  "LOAD_WORK_ORDER_DETAILS_PAGE",
+  "OPEN_LOCATION_EDITOR",
+] as const;
+
+function parseMessageEventData(data: unknown): unknown {
```
johnclary, line 11: 👍

signs-work-order-map/utils/iFrameMessenger.tsx
```diff
@@ -2,22 +2,56 @@
...
+      return JSON.parse(data);
+    } catch {
+      return null;
+    }
+  }
+  if (typeof data === "object") return data;
+  return null;
+}
+
+function isKnackToIFrameMessage(value: unknown): value is KnackToIFrameMessage {
```
johnclary, line 24: nice [type predicate](https://www.typescriptlang.org/docs/handbook/advanced-types.html#using-type-predicates)!

johnclary (approved):
Solid upgrade here—thanks for making this change. It's great to see all this data in sync 🚀

roseeichelmann (approved):
followed the test steps, this works as expected! could we maybe make the points the same size across blue/yellow? or is that a purposeful decision to make them different sizes?

mateoclarke:
Hey @roseeichelmann, good catch on the dot sizing. The mismatch wasn't intentional, but I was able to get the blue dots and yellow/green dots to scale at the same rate, depending on the zoom. They looked different because the blue dots are imported from AGOL and drawn by Mapbox as a circle layer and scale up as we zoom in so they're easier to tap on iPad. The yellow/green dots are separate HTML makers that had a fixed pixel size, so they didn't grow with zoom. I kept the touch-friendly blue size scaling and made the Knack markers use the same zoom-based sizing applied inline instead of as fixed styles.

signs-work-order-map/utils/iFrameMessenger.tsx
mddilley: very cool TIL for me 📚🙏

mddilley (approved):
All the test steps worked for me, and I see the dot sizing consistent across the colors as well. Glad that I checked back in on this work today! It is looking so good! 😎 🗺️

frankhereford (approved):
Worked through the test steps and everything appears to be acting correctly, and your code reads really well, especially the typescript aspects of it -- it helps me understand what's going on a lot. Thanks for this work! 🚢 🚢 🚢

